PR #25, opened 2021-06-14
refactor: update patchProps method to pass all props not only iterable ones

ftonato:
**Description:**
We currently are only iterate through the iterable props but we should iterate over all props...

**Status**

### DIFF
--- a/src/patch.ts
+++ b/src/patch.ts
@@ -9,14 +9,12 @@ import { VElement, VNode, VProps } from './m';
  * @param {VProps} newProps - New VNode props
  */
 export const patchProps = (el: HTMLElement, oldProps: VProps = {}, newProps: VProps = {}): void => {
-  if (!oldProps && !newProps) return;
-
   const oldPropKeys = Object.keys(oldProps);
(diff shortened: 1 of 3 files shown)


src/patch.ts
```diff
@@ -10,14 +10,12 @@ import { VElement, VNode, VProps, VFlags } from './m';
  */
 /* istanbul ignore next */
 export const patchProps = (el: HTMLElement, oldProps: VProps = {}, newProps: VProps = {}): void => {
-  if (!oldProps && !newProps) return;
```
ftonato, line 13: As `oldProps` and `newProps` are by default objects with specific types we never will get **Falsy** values for them, so we never will run that `return`

aidenybai:
I think in our case it's very similar and it might be smarter to use `for..of` because we `Object.keys` and then access on demand with `obj[prop]`. I think I switched back to `for..of` in a newer commit but I'm not sure whether there is a major performance impact on benchmarks. The main issue was shallow cloning of the object keys with `[...obj]`.

Issue #107, opened 2021-08-08
Question: patch performance.


hyrious:
The [patch algorithm with no _delta_ info](https://github.com/aidenybai/million/blob/372294f7088c7f65bd17d9afc37677823d8e763d/src/patch.ts#L80-L87) is quite simple and it's easy to make a worst case.

e.g. If we patch `A B C` &rarr; `X A B C`, this algorithm would compare & patch old_C with B, then old_B with A, then old_A with X, then insert C (4 steps). While the best way to do this is one step: insert X before old_A.

Better algorithms can be found in [inferno](https://github.com/infernojs/inferno/blob/master/packages/inferno/src/DOM/patching.ts), [solid](https://github.com/solidjs/solid/blob/main/packages/solid/store/src/modifiers.ts), [fre](https://github.com/yisar/fre/blob/master/src/reconcile.ts) and [vue](https://github.com/vuejs/vue-next/blob/master/packages/runtime-core/src/renderer.ts). (The solid.js one should be easiest to read.)

You may argue that using _delta_ is a better choice, it's the same to say we left a blank in dom diffing to let other programmers write. Besides, if we're making some ui framework which can analyse _template_ like vue or svelte, then the delta info can be cached and the performance is the best. -- only if we have _template_. If we choose to work like react, whose layout is determined at runtime (the render() func), then we get no way to pre calculate the delta, neither caching it.

Would you implement a better patch algorithm, or leave it as is?

aidenybai:
Good point, I've been trying to work on this issue but my knowledge + experience with algorithms is too little to pull something off like this.

I think deltas are still a viable feature, as it can be very useful, but I also agree with you the current patch algorithm can be refactored to be significantly more performant.

I've tried [to copy over and monkey-patch the diffing algorithm](https://github.com/aidenybai/million/blob/b431313022dbb7efc7fe269a1c0623d1a6a923b9/src/patch.ts#L164) from snabbdom, but only for keyed diffing. I chose this because I think generally if you want something to be performant, you should generally just use keys (if deltas don't apply). Also, it is somewhat difficult to determine whether two vnodes without a keys are equal without deep comparision (also not performant).

The snabbdom copied solution was unsuccessful, as when I ad-hoc benchmarked it, it was much slower than even worst case patching. 

I would love to implement a better patch algorithm, but my current skills and understanding on how to isn't there to figure out how to implement it correctly. I'll definetely take a look at the resources you provided, and if you want to help, feel free to if you have a bit of time.

[...]

PR #109, opened 2021-08-09
Efficient keyed diffing

aidenybai:

#107 

**Status**


loteoo:
That was fast 👏! Out of curiosity wouldn't non-keyed diffing be mostly the same thing but with just the "tag" as the key? It would be nice if it was possible to do both key and non-keys with the same piece of code 🤞 (this is all way over my head sorry if it doesn't make sense haha)

aidenybai:
It's somewhat difficult to do that because you can have two nodes with the same tag but with vastly different props and/or children. For example: 

```jsx
<div id="foo"></div>

<div class="bar baz">Hello World!</div>
```

Have exactly the same tag, but different attributes/children. The only viable way is if you diff the tag (constant), props (linear), children (recursive), flags (constant), and deltas (linear) for each node, and that could be really computationally expensive

aidenybai:
Current Bug: [A, B] being double swapped ([A, B] -> [B, A] -> [A, B]), no modification

Issue #191, opened 2022-06-15
bug: `Cannot set property form of #<HTMLButtonElement> which has only a getter`


gustavopch:
**Describe the bug**
When trying to create a button element with the `form` attribute, the following error happens:
```
Uncaught TypeError: Cannot set property form of #<HTMLButtonElement> which has only a getter
```

**To Reproduce**
1. Open https://stackblitz.com/edit/mega-millions-ryy3r5?file=index.html
2. See the error in the console

**Expected behavior**
It should be possible to create a button element with the `form` attribute just as it's possible to do this:
```
document.body.innerHTML = '<button form="some-form">OK</button>'
```

Maybe the `form` attribute (and possibly others too) can only be set with `setAttribute()`?

**Screenshots**
![image](https://user-images.githubusercontent.com/10941958/173918610-0b314edb-ba95-450c-ae23-1e3596e81fbb.png)

**Device (please complete the following information):**
 - OS: macOS
 - Browser: Chrome
 - Version: 102

**Additional context**
https://github.com/aidenybai/million/blob/74d50391113abad6c89d98eb3e93b748840ab51a/src/million/createElement.ts#L35-L37

aidenybai:
Fixed in #192

PR #192, opened 2022-06-16
fixed: do not set readonly property.

cbbfcd:
fixed: [#191 ]

refer to: [hyperapp](https://github.com/jorgebucaran/hyperapp/blob/main/index.js#L85)

bug fixed.
**Status**

### DIFF
--- a/src/million/createElement.ts
+++ b/src/million/createElement.ts
@@ -32,7 +32,13 @@ export const createElement = (vnode?: VNode | null, attachField = true): DOMNode
         } else if (propName.charCodeAt(5) === COLON_CHAR) {
           el.setAttributeNS(XLINK_NS, propName, String(propValue));
         }
-      } else if (el[propName] !== undefined && !(el instanceof SVGElement)) {
+      } else if (
+          el[propName] !== undefined && 
+          !(el instanceof SVGElement) &&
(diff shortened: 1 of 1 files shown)


src/million/createElement.ts
```diff
@@ -32,7 +32,13 @@ export const createElement = (vnode?: VNode | null, attachField = true): DOMNode
         } else if (propName.charCodeAt(5) === COLON_CHAR) {
           el.setAttributeNS(XLINK_NS, propName, String(propValue));
         }
-      } else if (el[propName] !== undefined && !(el instanceof SVGElement)) {
```
aidenybai, line 35: There is a bit more elegant solution. Because both list and form are `null` value instead of `undefined`, you can check the `null` case like this instead of hard code:

```js
el[propName] !== undefined && 
el[propName] !== null && 
!(el instanceof SVGElement) &&
propName in el
```

aidenybai: Could you also make this change in `useProps.ts`?

aidenybai (approved):

aidenybai:
LGTM, thank you!

src/million/createElement.ts
cbbfcd: ok~

PR #198, opened 2022-06-19
fix(createElement): ignore prop value with null and undefined

SukkaW:

**Status**

I am using JSX with babel automatic runtime transformation and got `Uncaught TypeError: setting getter-only property "children"`. It is caused by `jsx` function that set `props.children` to `undefined`:

https://github.com/aidenybai/million/blob/e0896e69ecbed516c7e2c77dbe8c0fe7fca0a32e/src/jsx-runtime/jsx.ts#L12

However, `createElement` will always set prop (including `children`) even if the prop is read-only and the prop value is `undefined`. The PR adds an additional `null` and `undefined` check before applying props. The corresponding test case has also been added as well.

### DIFF
--- a/src/million/createElement.ts
+++ b/src/million/createElement.ts
@@ -24,15 +24,17 @@ export const createElement = (vnode?: VNode | null, attachField = true): DOMNode
         } else if (propName.startsWith('xlink')) {
           el.setAttributeNS(XLINK_NS, 'href', String(propValue));
         }
-      } else if (
-        el[propName] !== undefined &&
-        el[propName] !== null &&
-        !(el instanceof SVGElement) &&
-        propName in el
-      ) {
-        el[propName] = propValue;
-      } else {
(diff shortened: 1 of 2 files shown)


aidenybai:
Looks good, could you also make this change in `useProps.ts` as well?

SukkaW:
> Looks good, could you also make this change in `useProps.ts` as well?

I have checked `useProps`, it has already included a prop value check:

https://github.com/aidenybai/million/blob/e0896e69ecbed516c7e2c77dbe8c0fe7fca0a32e/src/million/drivers/useProps.ts#L47

aidenybai:
Cool, thanks for the help! Will be pushed out in the next patch

aidenybai:
Hey! I seen you've made some PR's to the Million repo, If you'd be willing I'm extending an invitation for you to become a collaborator :)

Link: https://github.com/aidenybai/million/invitations

PR #204, opened 2022-06-25
feat: exclude some properties, like children

cbbfcd:
avoid setting unwanted properties to element nodes.

image

**Status**

### DIFF
--- a/.husky/config-msg
+++ b/.husky/config-msg
@@ -1,4 +1,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-npx --no-install commitlint --edit "$1"
--- a/package.json
+++ b/package.json
@@ -115,14 +115,8 @@
     "url": "https://github.com/aidenybai/million/issues"
   },
   "homepage": "https://github.com/aidenybai/million",
-  "commitlint": {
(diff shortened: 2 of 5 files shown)


aidenybai:
Thoughts on having a one-liner here: https://github.com/aidenybai/million/blob/d0cb29835a282c18e7ab58a83ebccb008a11b1c2/src/million/m.ts#L96

`if (props?.children) delete props.children;` Like this so that children is never passed through during VNode creation?

cbbfcd:
ehhh~, It is indeed a more reasonable solution, let me try.
btw, i try to add some unit-test for this `pr`, but [there a bug for vitest now](https://github.com/vitest-dev/vitest/issues/1537), 😂 wait for `vitest` team to fix this

aidenybai:
> ehhh~, It is indeed a more reasonable solution, let me try. btw, i try to add some unit-test for this `pr`, but [there a bug for vitest now](https://github.com/vitest-dev/vitest/issues/1537), 😂 wait for `vitest` team to fix this

Gotcha, thanks for the flexibility! Lets hope Vitest team fixes soon 😂😂😂, feel free to merge PR yourself when you finish.

PR #230, opened 2022-08-01
fix: `welcome` command not working on initial run

QuiiBz:

When running the `welcome` script for the first time, I've noticed that the command fails with the following error:

Screenshot 2022-08-01 at 10 21 23

Digging a bit into it, I found that `dev/script.tsx` does not exist because the `pnpm run dev` command is not awaited - but I understand why it's not because the command runs `vite` and such never ends unless we exit the process.

So I've moved the file writing to `dev/script.tsx` from the `welcome` script to the `dev` one. For reference, this error was thrown only if we don't have the `dev/script.tsx` file - so only on the initial run of this command (= if we don't have the `dev` folder).

Also, not really sure if it's intentional to update the dependencies when committing (ESLint has been bumped to 8.21.0, tell me if I need to revert this)

**Status**

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "bumpp": "^8.2.1",
     "c8": "^7.12.0",
     "esbuild": "^0.14.51",
-    "eslint": "^8.20.0",
+    "eslint": "^8.21.0",
     "eslint-config-prettier": "^8.5.0",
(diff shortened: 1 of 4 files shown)


aidenybai:
Package bumping should be intentional! LGTM :)

PR #238, opened 2022-08-03
feat(react): add `componentDidCatch`

### LINKED ISSUE #222: feat: Error boundaries
https://github.com/aidenybai/million-react/issues/6

QuiiBz:

Closes #222
Relates to https://github.com/aidenybai/million-react/issues/6

In draft.

Currently, `componentDidCatch` in class components only works for render errors, not for effects. It's a lot harder to achieve considering the current architecture of Million, because we need to retrieve recursively the parent VNode of a VNode, until is a class component and has `componentDidCatch`.

**UPDATE**
I found a way to catch all errors from effects, but it might add a (lot?) of overhead. In each VNode, we have to keep track of the parent VNode (`_parent`). Class Components also add a `_component` field to the root VNode. When an error is thrown from an effect (initial or subsequent), we traverse up the three of VNodes until we find one that has a `_component`, and run its `componentDidCatch`. This is similar to Preact's implementation.

**Status**


aidenybai:
This looks like the best solution for classes. LGTM, merge when you're ready!

QuiiBz:
I'll research a bit more about catching errors from effects before merging, because `componentDidCatch` currently only works for (re)render errors.

QuiiBz:
Found an issue with the changes inside the Vite plugin, it was throwing an error and such fail to transform the AST.

Fixed by only adding the `_parent`  to the root velement of a class/function component instead of every velement (previously done in `m`), and remove the faulty code inside vite-plugin-million.

Left: this PR
Right: main branch

Screenshot 2022-08-06 at 08 58 12

packages/million/render.ts
```diff
@@ -22,6 +22,9 @@ export const patch = (
   hook: Hook = () => true,
   effects: Effect[] = [],
 ): DOMNode => {
+  if (typeof newVNode === 'object' && typeof oldVNode === 'object') {
+    newVNode._parent = oldVNode._parent;
```
aidenybai, line 26: opt in logic?

aidenybai:
Congrats on the ship!!

PR #250, opened 2022-08-04
fix(compiler): handle literal attribute keys

jasonappah:

Once an attribute key includes a special character like a hyphen, its type is `Literal` instead of `Identifier`, as other alphanumeric keys would have. Literals don't have the name property for getting the key's value like Identifiers do. This causes attributes with non-alphanumeric keys like ARIA properties to be rendered to the DOM with an undefined key, which is fixed by the PR.

**Status**


aidenybai:
Great work, LGTM, merge when you're ready :)

PR #254, opened 2022-08-10
fix: new implementation of useSyncExternalStore

aidenybai:

**Status**

  - x ] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)


packages/react/hooks.ts
```diff
@@ -232,11 +232,14 @@ export const useDeferredValue = (value: any) => {
...
-  const state = useState(getSnapshot());
+  const [, updateState] = useState();
+  const forceUpdate = useCallback(() => updateState({}), []);
+  const [subscriber, setSubscriber] = useState(() => subscribe(forceUpdate));
+
   useEffect(() => {
-    subscribe(state);
-  });
-  subscribe(state);
+    if (subscriber) subscriber();
```
QuiiBz, line 240: Can be simplified and maybe save some bytes?
```suggestion
    subscriber?.();
```

QuiiBz (approved):
One small suggestion, otherwise LGTM!

SukkaW (changes requested):
The implementation still seems wrong.

IIRC in react's implementation, `useSyncExternalStore` will unsubscribe and then re-subscribe if `subscribe` function changes, and bail out re-subscribe if `subscribe` doesn't change.

So it is not the `getSnapshot()` but the `subscribe` that should be in the `useEffect`'s dependency array.

Also, some other existing implementation that we could learn from:

- preact's implementation: https://github.com/preactjs/preact/blob/6b92b1fab41599e6da4f96d65b07fdbe0b6ff2fc/compat/src/index.js#L136-L154
- react's `use-sync-external-store` polyfill implementation: https://github.com/facebook/react/blob/main/packages/use-sync-external-store/src/useSyncExternalStoreShimClient.js

packages/react/hooks.ts
```diff
@@ -232,11 +232,14 @@ export const useDeferredValue = (value: any) => {
...
+  const [subscriber, setSubscriber] = useState(() => subscribe(forceUpdate));
+
   useEffect(() => {
-    subscribe(state);
-  });
-  subscribe(state);
+    subscriber?.();
+    setSubscriber(subscribe(forceUpdate));
+  }, [getSnapshot()]);
   return state;
```
SukkaW, line 243: Where does the `state` come from if `const state = useState(getSnapshot());` is removed?

aidenybai: I think we can go ahead with copying Preact's implementation.

aidenybai:
I see, thank you for the helpful explanation @SukkaW. I think we should go ahead and copy Preact in this case (thank you preactjs!) to be safe.

QuiiBz:
> ...copy Preact in this case...

Let's not forget to credit them inside the file :)

aidenybai:
good catch, adding attribution

SukkaW (approved):

PR #258, opened 2022-08-13
fix: wouter compat

### LINKED ISSUE #203: bug: Wouter compat
**Describe the bug**
Wouter errors when integrated into a Million/React environment

**To Reproduce**

```js
import { createRoot, useState } from 'million/react';
import { Link, Route } from 'wouter';

function App() {
  const [count, setCount] = useState(0);
  return (
    <div>
      <button onClick={() => setCount(count + 1)}>{count}</button>

      <Link href="/users/1">
        <a className="link">Profile</a>
      </Link>

      <Route path="/about">About Us</Route>
      <Route path="/users/:name">{(params) => <div>Hello, {params.name}!</div>}</Route>
    </div>
  );
}

createRoot(document.body).render(<App />);
```

Gives something like this:

```
wouter.js?v=b667f2ca:1129 Uncaught TypeError: Cannot read properties of null (reading 'i')
    at useRef (wouter.js?v=b667f2ca:1129:14)
    at useNavigate (wouter.js?v=b667f2ca:1636:18)
    at Link (wouter.js?v=b667f2ca:1659:18)
    at compat.ts:15:17
    at hook2 (hooks.ts:37:17)
    at Object.createComponent [as handleFunction] (compat.ts:39:4)
    at Object.h (h.ts:45:14)
    at App (script.tsx:15:19)
    at compat.ts:15:17
    at hook2 (hooks.ts:37:17)
``` 

**Expected behavior**

Should behave as [Wouter](https://github.com/molefrog/wouter) intends it to

QuiiBz:

Fix #203, WIP

Two issues to achieve wouter compat:

1. In Million, in a `VElement`, `children` is always an array of `VNode`. But in React, `children` can be a `VNode` or an array of `VNode`. Here are 3 components to verify this: 
image
Million:
image
React:
image
wouter's `Link` uses `isValidElement` and `cloneElement`, so this PR makes them both accept and handle correctly vnode and array of vnode.

2. `cloneElement` is used inside wouter's `Link`, and passes a second parameter `extraProps` that wasn't supported by Million:
image
As we can see in React and as described in the [docs](https://reactjs.org/docs/react-api.html#cloneelement), this param contains new props to add to the cloned element: https://github.com/facebook/react/blob/d5f1b067c8bbb826b823d0354a28ba31078b70c0/packages/react/src/ReactElement.js#L486
This PR add this new parameter in `cloneElement` (I haven't yet added the third one, which is a rest of children).

**Status**

### DIFF
--- a/packages/react/compat.ts
+++ b/packages/react/compat.ts
@@ -25,15 +25,13 @@ const catchError = (vnodeLike: { _component?: Component, _parent?: VElement } | 
 }
 
 const addParentToChildren = (velement: VElement) => {
-  if (Array.isArray(velement.children)) {
-    velement.children?.forEach(child => {
-      // if child is null, `typeof child === "object"` still
-      // returns true so we have to check if child is defined
-      if (child && typeof child === 'object') {
-        child._parent = velement
-      }
-    })
-  }
+  velement.children?.forEach(child => {
+    // if child is null, `typeof child === "object"` still
+    // returns true so we have to check if child is defined
+    if (child && typeof child === 'object') {
+      child._parent = velement
+    }
+  })
 }
 
 export const getCircularReplacer = () => {
--- a/packages/react/compat.ts
+++ b/packages/react/compat.ts
@@ -25,11 +25,11 @@ const catchError = (vnodeLike: { _component?: Component, _parent?: VElement } | 
 }
 
 const addParentToChildren = (velement: VElement) => {
-  velement.children?.forEach(child => {
-    if (typeof child === 'object') {
-      child._parent = velement
-    }
-  })
+  // velement.children?.forEach(child => {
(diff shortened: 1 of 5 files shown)


packages/react/compat.ts
```diff
@@ -26,7 +26,9 @@ const catchError = (vnodeLike: { _component?: Component, _parent?: VElement } |
 
 const addParentToChildren = (velement: VElement) => {
   velement.children?.forEach(child => {
-    if (typeof child === 'object') {
+    // if child is null, `typeof child === "object"` still
+    // returns true so we have to check if child is defined
+    if (child && typeof child === 'object') {
```
cbbfcd, line 31: 👍

cbbfcd (approved):

PR #286, opened 2022-11-21
fix(#286): memo working incorrectly

### LINKED ISSUE #285: bug: React.memo works incorrectly
**Describe the bug**
Hey. I'm trying to test the compatibility layer with react. I came across a bug when using the memo function. This function is not properly passing props. I think the point is that memo function is implemented by passing props through `Object.values(props) `which produces an array instead of an object. As a result, the memoizable component receives an array of arguments as input, instead of a single props argument.

**To Reproduce**
Wrap any component in memo and pass some props to it.

**Expected behavior**
Props must be passed as a single argument.

aidenybai:
Fixes #285


packages/react/react.ts
```diff
@@ -48,13 +50,15 @@ const isValidElement = (vnode?: VNode | VNode[] | null) => {
...
 };
 
-const memo = (
-  component: (...args: unknown[]) => VNode
-) => {
+const memo = (component: (...args: unknown[]) => VNode) => {
+  const cache = new Map();
   return (props: VProps) => {
-    return component(Object.values(props));
+    if (!cache.has(props)) return cache.get(props);
```
SukkaW, line 56:
```suggestion
    if (cache.has(props)) return cache.get(props);
```

Also IMHO the `React.memo` is using shallow compare by default (while `Map.prototype.has` is using `SameValueZero` which compares the reference equality).

SukkaW (changes requested):

packages/react/react.ts
atellmer: This implementation of memoization will create a memory leak because there is no mechanism for clearing the cache. React also memoizes only the last render.

aidenybai: I see, thank you for the feedback. Let me adjust the code to fix

packages/react/react.ts
```diff
@@ -48,13 +50,17 @@ const isValidElement = (vnode?: VNode | VNode[] | null) => {
...
 
-const memo = (
-  component: (...args: unknown[]) => VNode
-) => {
+const memo = (component: (...args: unknown[]) => VNode) => {
+  let prevProps;
+  let prevRet;
   return (props: VProps) => {
-    return component(Object.values(props));
+    if (prevProps === props) return prevRet;
```
SukkaW, line 57: This won't work as intended.

Consider this:

```jsx
/** @jsxRuntime automatic */
/** @jsxImportSource million */

const Component = memo((props) => {
  return <></>;
});

<Component props1="Hello, Million.js!" />
```

The JSX above will be compiled into:

```
import { jsx as _jsx, jsxs as _jsxs, Fragment as _Fragment } from "million/jsx-runtime";
const Component = memo((props)=>{
    return _jsx(_Fragment, {});
});

_jsx(Component, {
    props1: "Hello, Million.js!"
});
```

As you can see, the props (`{ props1: "Hello, Million.js!" }`) is a plain object. When re-rendered, `prevProps === props` will always return false.

SukkaW (changes requested):

packages/react/react.ts
aidenybai: This is very helpful and thanks for being patient with me. Let me update the code and run it by you again

packages/react/react.ts
```diff
@@ -50,11 +50,19 @@ const isValidElement = (vnode?: VNode | VNode[] | null) => {
...
 
+// taken from preact/compat
+// https://github.com/preactjs/preact/blob/4a37c998833dafa42bcdb5250128d76ee3307414/compat/src/util.js#L19
+const arePropsDifferent = (a: VProps, b: VProps) => {
+  for (const i in a) if (!(i in b)) return true;
+  for (const i in b) if (a[i] !== b[i]) return true;
+  return false;
+};
+
 const memo = (component: (...args: unknown[]) => VNode) => {
```
SukkaW, line 61:
```suggestion
const memo = (component: (...args: unknown[]) => VNode, compare = arePropsDifferent) => {
```

packages/react/react.ts
```diff
@@ -50,11 +50,19 @@ const isValidElement = (vnode?: VNode | VNode[] | null) => {
...
   let prevProps;
   let prevRet;
   return (props: VProps) => {
-    if (prevProps === props) return prevRet;
+    if (!arePropsDifferent(props, prevProps)) return prevRet;
```
SukkaW, line 65:
```suggestion
    const shouldComponentUpdate = isEqual
      ? shouldComponentUpdate = !isEqual(props, prevProps)
      : arePropsDifferent(props, prevProps);

    if (!shouldComponentUpdate) return prevRet;
```

packages/react/react.ts
SukkaW: `React.memo` accepts a second argument for the custom `isEqual` function.

packages/react/react.ts
atellmer: In this case, the closure will not work as expected and prevProps will be the same for all instances of the memoized component. Because the memo function is called once when the memo component is created. You need to store previous props the same way you store state in useState , at the component instance level.

PR #291, opened 2023-01-10
fix: implementation of useSyncExternalStore

exuanbo:

**Status**

This PR updates the hook implementation used by `preact`.

Refs: https://github.com/preactjs/preact/pull/3663 and https://github.com/preactjs/preact/commit/528a7760c1cdec7515d514c71875860528e6dc12


packages/react/hooks.ts
```diff
@@ -243,26 +244,37 @@ export const useDeferredValue = (value: any) => {
...
   const value = getSnapshot();
 
-  useLayoutEffect(() => {
-    if (value !== state) {
-      setState(() => value);
-    }
-  }, [subscribe, value, getSnapshot]);
+	const [{ instance }, forceUpdate] = useState({
+		instance: { value, getSnapshot }
+	});
```
SukkaW, line 254: It looks like preact's new `useSyncExternalStore` is now ported from react's `use-sync-external-store/shim`. So you should do it too.

Please do copy their comments. Million.js doesn't have concurrent rendering (Both preact and Million.js performs synchronous rendering just like React pre-18), and it is the only reason why react's shim also works for Million.js (and preact).

https://github.com/facebook/react/blob/main/packages/use-sync-external-store/src/useSyncExternalStoreShimClient.js

SukkaW (approved):

SukkaW:
Thanks for your contribution! I have just approved the CI.

PR #292, opened 2023-02-16
v2.0.0-beta

aidenybai:

The Block API is intended to be used as an implementation detail (either under runtime abstraction or compiler abstraction).

One idea I have for this API is to create a <Million>user code</Million> component for React and Preact. This allows Million to have fast impact and easy integration whenever the developer requires faster rendering. Personally, this is a big advantage (compat libraries that Million, Inferno, and Preact have all have to try to maintain compat), but with this idea, it's just maintaining a React component.

The second is to build a compiler upon this API. A React component (or any component) can be compiled into blocks, allowing for very fast perf.

See https://github.com/aidenybai/million/tree/refactor-core/packages/block#readme

**Benchmark setup for reference:**

1. Clone https://github.com/aidenybai/js-framework-benchmark
2. Setup binaries for benchmarking: https://github.com/aidenybai/js-framework-benchmark#1-new-run-pre-built-binaries-for-all-frameworks
3. Make sure you have three terminal instances running
- 1. Start dev server: `npm run start`
- 2. Build the bundle for Million:
```
cd frameworks/keyed/million
npm i
npm build-dev
```
Make sure you modify the path to your `packages/block/index.ts` here: https://github.com/aidenybai/js-framework-benchmark/blob/31f620323c8e9e306937fcc6f1aca7c17f3c3283/frameworks/keyed/million/src/main.js#L5
- 3. Run benchmark:`npm run bench -- --framework keyed/million --benchmark 01_ 02_ 03_ 04_ 05_ 06_ 07_ 08_ 09_ && npm run results`

4. Then open the `./webdriver-ts-results/table.html` file

TODO:
- [x] tests
- [x] react compat
- [x] keep legacy react compat + legacy compiler, remove other packages
- [x] create new react component
- [x] remove legacy benchmarks

**Status**

  - [x] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)


packages/block/README.md
```diff
@@ -0,0 +1,48 @@
...
+## Example Usage
+
+Inside the `/dev` folder, put the following code in `index.js`:
+
+```js
+import { createBlock, fragment } from 'packages/block';
+
+const block = createBlock((props) => {
+  // Any prop within props isn't the actual value, it's a "Hole"
+  // always interpolate into the JSX.
```
SukkaW, line 21: Is that mean we are no longer able to pass the event handlers (functions) to the `block`?

aidenybai: You can pass event handlers like:

```js
const block = createBlock(({ listener, value }) => <div onClick={listener} onMouseOver={() => alert('hi!')}>{value}</div>)
```

But this is not allowed:

```js
const block = createBlock(({ listener, value }) => <div onMouseOver={() => listener(params)}>{ value + 1 }</div>)
```

packages/block/block.ts
```diff
@@ -0,0 +1,220 @@
...
+export const HOLE_PROXY = new Proxy(
+  {},
+  {
+    // A universal getter will return a Hole instance if props[any] is accessed
+    // Allows code to identify holes in virtual nodes ("digs" them out)
+    get(_, prop: string): Hole {
+      if (mapHas$.call(holeCache, prop)) return mapGet$.call(holeCache, prop);
+      const hole = new Hole(prop);
+      mapSet$.call(holeCache, prop, hole);
+      holeCache.set(prop, hole);
```
SukkaW, line 30: Is it possible to leak memory here? It is possible to have infinite props, so the Map size will grow indefinitely.

aidenybai: good point, but it is also good to have a cache here (instantiating a lot of holes can be wasteful), any ideas for a good middle ground?

packages/block/template.ts
```diff
@@ -0,0 +1,124 @@
...
+  for (let name in vnode.props) {
+    const value = vnode.props[name];
+    if (name === 'key' || name === 'ref' || name === 'children') {
+      continue;
+    }
+    if (name === 'className') name = 'class';
+    if (name === 'for') name = 'htmlFor';
+
+    // eslint-disable-next-line @typescript-eslint/prefer-string-starts-ends-with
+    if (name[0] === 'o' && name[1] === 'n') {
```
SukkaW, line 29: We should do a performance benchmark on this. On Chrome `startsWith` should be the fastest. [Benchmark](https://perf.link/#eyJpZCI6Im9xa2lrdTdjNWh1IiwidGl0bGUiOiIiLCJiZWZvcmUiOiJjb25zdCBzdHIgPSAnb25DbGljayciLCJ0ZXN0cyI6W3sibmFtZSI6InN0YXJ0c1dpdGgiLCJjb2RlIjoic3RyLnN0YXJ0c1dpdGgoJ29uJyk7IiwicnVucyI6WzQzNTQwMDAsMTUzNjAwMCw1MDQxMDAwLDU4ODYwMDAsMTMxMjAwMCw0NDEwMDAwLDE5MzQwMDAsMzQxOTAwMCw1MTM2MDAwLDE2NzAwMCw0NjY0MDAwLDIyMzUwMDAsNTQyNzAwMCw0NTI3MDAwLDUwNjQwMDAsNjUyMDAwLDM0NTUwMDAsNDc4MDAwMCw1MjEzMDAwLDQ5NTAwMCw2MTI0MDAwLDIyODAwMCwxMDgwMDAwLDQxODEwMDAsODEwMDAwLDMxMzYwMDAsMzUyMzAwMCwzMjA1MDAwLDEzMDAwMCwxNDQ3MDAwLDIyMzAwMCwxOTUwMDAsODkwMDAwLDMyNjQwMDAsMzIyMTAwMCwyOTY2MDAwLDM0NzAwMDAsMTE1MDAwLDQyMTIwMDAsMjA5ODAwMCw1NDI4MDAwLDE2ODYwMDAsMjA0NDAwMCw3NTEwMDAsMTg5NDAwMCwxMDc0MDAwLDI4NzYwMDAsMjc4NzAwMCw1NDQ4MDAwLDI4MzkwMDAsNDQ2MDAwLDE1MzcwMDAsMTY4ODAwMCw1Njc4MDAwLDgxODAwMCwzMDMwMDAsMTg2NzAwMCwxNzAyMDAwLDI3MDAwMCwyNTE5MDAwLDU2NDIwMDAsMzY5MjAwMCw1NDI4MDAwLDU2NjIwMDAsMzE1MzAwMCw1NDI4MDAwLDU0MjgwMDAsNTQyODAwMCwyMDYxMDAwLDU4ODYwMDAsMjk4ODAwMCw2MDkwMDAsMjkyNDAwMCwxNTM4MDAwLDQyNjgwMDAsNDIzNzAwMCwxNDQwMDAsNDIyNzAwMCwxNDU5MDAwLDIxMTEwMDAsNTIzODAwMCw4MTkwMDAsNTQyODAwMCwzOTk5MDAwLDMxNTIwMDAsMjMwMDAsMzUwMDAwLDU0MjgwMDAsNjAwODAwMCwzNzQyMDAwLDU0MjgwMDAsNTQyODAwMCwyODQzMDAwLDMwMDAwMCwxNTczMDAwLDUwMjIwMDAsMTE3OTAwMCwzMzQzMDAwLDMzMzAwMDAsNDI2MDAwXSwib3BzIjoyOTMyNTIwfSx7Im5hbWUiOiJjaGFyQ29kZUF0IiwiY29kZSI6InN0ci5jaGFyQ29kZUF0KDApID09PSAxMTEgJiYgc3RyLmNoYXJDb2RlQXQoMSkgPT09IDExMDsiLCJydW5zIjpbNDA4NDAwMCwxMzMyMDAwLDQwNzAwMCw0MjIwMDAwLDMyNzcwMDAsMjYzNjAwMCwzNjI4MDAwLDUzMTkwMDAsNDkxMTAwMCwyNDg2MDAwLDgyNjAwMCwxNzY1MDAwLDMyNzcwMDAsMTM5MzAwMCwyNTk5MDAwLDMyNzYwMDAsMjM0MDAwLDQ3MzAwMCw0NDg0MDAwLDM1NjAwMDAsMjA5NTAwMCwzNjI5MDAwLDI1MzAwMCwxMzEzMDAwLDMyNzcwMDAsMzUxOTAwMCw0NTAzMDAwLDE2NDMwMDAsMTY0NTAwMCwzMjc3MDAwLDM4NjEwMDAsMjUxNTAwMCwzNDc0MDAwLDE3NTgwMDAsMTgxMTAwMCwxNDkwMDAwLDg2MzAwMCwzNDYzMDAwLDUwNjQwMDAsMzgwNzAwMCw0NzczMDAwLDkzMDAwLDE4MDAwLDE3MTYwMDAsNDEwMzAwMCw0MTE5MDAwLDE5ODcwMDAsMjk0ODAwMCwzMjc3MDAwLDE3NjAwMCwxNjk5MDAwLDQwMDAsMjA3NTAwMCwxNjIxMDAwLDM2MzMwMDAsMTg2MjAwMCwxMzg0MDAwLDI0NzgwMDAsMzMxMTAwMCw0MTkyMDAwLDMyNzcwMDAsMTUwMDAsNTU3MDAwLDI2MDgwMDAsMjUxNjAwMCwzMzYzMDAwLDE4NTMwMDAsMzQ4NDAwMCwzMjc2MDAwLDExMTMwMDAsMTM3MjAwMCwyMjQ5MDAwLDEyNDUwMDAsMjE3OTAwMCw1MjY3MDAwLDI2MDEwMDAsMjQ3MDAwMCw1MTQwMDAsNTIyOTAwMCwzNDczMDAwLDQwOTkwMDAsMzI3NzAwMCwzNDQxMDAwLDMxMDQwMDAsMjc3OTAwMCwxMDUxMDAwLDMzNDgwMDAsOTg2MDAwLDI4NjIwMDAsMzI3NzAwMCwzMjc3MDAwLDQzODAwMCwxMDk3MDAwLDI0MTcwMDAsNTUzMDAwLDMyNzcwMDAsNDUzNDAwMCwxMDUxMDAwLDE0MzAwMDAsMjk1MDAwMF0sIm9wcyI6MjUwODI1MH0seyJuYW1lIjoiWzBdIGFuZCBbMV0iLCJjb2RlIjoic3RyWzBdID09PSAnbycgJiYgc3RyWzFdID09PSAnbic7IiwicnVucyI6WzE5ODAwMDAsMTQ0NjAwMCwzMDcyMDAwLDI1MjEwMDAsNTYwNDAwMCw0MDUzMDAwLDQwNTMwMDAsMTEyMDAwMCw2MjYwMDAsMzE4MDAwMCwxOTI4MDAwLDQwNzgwMDAsMTcwMDAsNTQyNTAwMCwyOTc5MDAwLDI5NDAwMCwxODk1MDAwLDIxNDIwMDAsNjMwMDAsMzI4MzAwMCw0OTY2MDAwLDIyMTAwMCwzMTQzMDAwLDgyNDAwMCw1MzY4MDAwLDE3MzIwMDAsMjcyNTAwMCwyOTAwMCwyMDY4MDAwLDM0NDIwMDAsMjIwMTAwMCwzODMyMDAwLDUxMjkwMDAsNTIxMDAwMCw0OTE1MDAwLDY3NTAwMCw0Mjc2MDAwLDI4NzAwMCw0ODczMDAwLDIxNjMwMDAsMTcwMDAsMzg0NzAwMCwzMjIxMDAwLDM5MDcwMDAsMTQ0NjAwMCwzMTkzMDAwLDQ5NjAwMCw1Mzg5MDAwLDM1MDQwMDAsNDk1NjAwMCw0Mjc3MDAwLDQyNjMwMDAsNTI1ODAwMCw0MDUzMDAwLDQxMDIwMDAsMTc2OTAwMCw1NDAwMCwzMjQwMDAsNDA1MzAwMCwyOTcwMDAwLDE0NjgwMDAsMTgyMjAwMCwzNjczMDAwLDQwNTMwMDAsOTIyMDAwLDQ1NDYwMDAsNDg2MzAwMCw0MjAwMDAwLDM3OTkwMDAsNDA1MzAwMCwyOTM4MDAwLDg3MDAwMCw5MDQwMDAsMTA0ODAwMCwxODYxMDAwLDM1NzAwMDAsMTY1ODAwMCwyMTQ2MDAwLDM0NzkwMDAsMTE3NDAwMCw3OTQwMDAsMzE5ODAwMCwxNzUyMDAwLDU1NTAwMCw1MzYyMDAwLDMzNDUwMDAsMzU0NzAwMCw0MDUyMDAwLDQwNTMwMDAsMTMwNDAwMCwyNTg5MDAwLDMzNTUwMDAsMTUxNzAwMCw0MzgwMDAsNDA1MzAwMCwzMTU1MDAwLDQwNTIwMDAsNTQwMDAsMzEwMDAsNDM3NDAwMF0sIm9wcyI6MjczNTY0MH1dLCJ1cGRhdGVkIjoiMjAyMy0wMi0xNlQwOTowNjoxMy4yNjNaIn0%3D)

packages/block/block.ts
SukkaW: > any ideas for a good middle ground?

I have some rough ideas, but they all have drawbacks:

- limit size cache or even a limit size LRU cache to prevent memory leak
- the memo (deep equal -> referential equality) to prevent infinite props

SukkaW: I just came up with an example:

```ts
const block = createBlock(/** */);
for (let i = 0; i < 1000000; i++) {
  block({ title: 'Hello World' });
}
```

Here is the problem, although the value is always the same, the referential equality doesn't retain as the props are created as an object literal (`Object.is({ title: 'Hello World' }, { title: 'Hello World' }) === false`). And it will cause a memory leak.

packages/block/template.ts
aidenybai: fixed!

packages/block/block.ts
SukkaW: I have an idea. We should not have a single `HOLE_PROXY` across the entire library.

We should create a `hole proxy` instance for every block when `createBlock` is invoked. So every block created will have a `hole proxy` instance bound to them.

And instead of having an infinite cache, a block only has to remember the last prop value (a cache with size 1). Only when the prop changes, the block will re-create a `hole proxy` instance.

packages/block/README.md
SukkaW: We might have a hard time teaching people about this. A common mistake could be:

```
const block = createBlock(({ handleStateUpdate, value }) => {
  const handler = () => handleStateUpdate(value + 1);
  return <div onMouseOver={handler}>{ value + 1 }</div>;
})
```

packages/block/template.ts
QuiiBz: Depending on the website, results aren't the same at all:

Screenshot 2023-02-16 at 10 43 44

Screenshot 2023-02-16 at 10 43 54

But agree on the benchmarks, we should have them on a repository so we can track the performance improvements in V8.

SukkaW: > Depending on the website, results aren't the same at all:
>
> But agree on the benchmarks, we should have them on a repository so we can track the performance improvements in V8.

Yes, that's the point.
But I do prefer `startsWith` here. Not only is it fast, but also a few bytes smaller.

packages/block/block.ts
aidenybai: Got it! Let me make that change

packages/block/README.md
aidenybai: This is a good point, but the Block API is intended to be used as an implementation detail (either under runtime abstraction or compiler abstraction). 

One idea I have for this API is to create a `<Million>user code</Million>` component for React and Preact. This allows Million to have fast impact and easy integration whenever the developer requires faster rendering. Personally, this is a big advantage (compat libraries that Million, Inferno, and Preact have all have to try to maintain compat), but with this idea, it's just maintaining a React component.

The second is to build a compiler upon this API. A React component (or any component) can be compiled into blocks, allowing for very fast perf.

packages/block/types.ts
```diff
@@ -0,0 +1,82 @@
...
+  cache?: Map<number, HTMLElement>;
+  patch(_block: AbstractBlock) {}
+  mount(_parent?: HTMLElement, _refNode: Node | null = null) {}
+  move(_block: AbstractBlock | null = null, _refNode: Node | null = null) {}
+  remove() {}
+  toString() {}
+  get parent(): HTMLElement | null | undefined {
+    return this._parent;
+  }
+}
```
QuiiBz, line 29: I think this class can be made `abstract`, along with the methods:

```suggestion
export abstract class AbstractBlock {
  root?: HTMLElement;
  edits?: Edit[];
  el?: HTMLElement;
  _parent?: HTMLElement | null;
  props?: Props | null;
  key?: string;
  cache?: Map<number, HTMLElement>;
  abstract patch(block: AbstractBlock): HTMLElement;
  abstract mount(parent?: HTMLElement, refNode: Node | null): HTMLElement;
  abstract move(block: AbstractBlock | null = null, refNode: Node | null): void;
  abstract remove(): void;
  abstract toString(): string;
  get parent(): HTMLElement | null | undefined {
    return this._parent;
  }
}
```

packages/block/block.ts
```diff
@@ -0,0 +1,222 @@
...
+export class Block extends AbstractBlock {
+  root: HTMLElement;
+  edits: Edit[];
+  constructor(
+    root: HTMLElement,
+    edits: Edit[],
+    props?: Props | null,
+    key?: string,
+  ) {
+    super();
```
QuiiBz, line 76: `super()` is useless here, as `AbstractBlock` doesn't have a constructor

```suggestion
```

packages/block/block.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    for (let i = 0, j = this.edits.length; i < j; ++i) {
+      const current = this.edits[i]!;
+      const el = getCurrentElement(current, root, this.cache, i);
+      for (let k = 0, l = current.edits.length; k < l; ++k) {
+        const edit = current.edits[k]!;
+        const hasHole = 'hole' in edit && edit.hole;
+        const value = hasHole ? this.props![edit.hole!] : undefined;
+
+        if (edit.type === EditType.Block) {
```
QuiiBz, line 97: It might be more efficient to use `if ... else if ... else if` or use a `switch` statement, because the JS engine will only have a single path to check

packages/block/template.ts
```diff
@@ -0,0 +1,123 @@
...
+        patch: undefined,
+        block: undefined,
+      });
+      continue;
+    }
+
+    if (value) props += ` ${name}="${String(value)}"`;
+  }
+
+  if (IS_VOID_ELEMENT.test(vnode.tag)) return `<${vnode.tag}${props} />`;
```
QuiiBz, line 62: I'm not sure if a regex is faster than e.g using an array of tags (`TAGS.includes(vnode.tag)`)

aidenybai: Seems like you're right https://perf.link/#eyJpZCI6ImxmNWNnZzRjZDkxIiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IElTX1ZPSURfRUxFTUVOVCA9IC9hcmVhfGJhc2V8YmFzZWZvbnR8Ymdzb3VuZHxicnxjb2x8Y29tbWFuZHxlbWJlZHxmcmFtZXxocnxpbWFnZXxpbWd8aW5wdXR8aXNpbmRleHxrZXlnZW58bGlua3xtZW51aXRlbXxtZXRhfG5leHRpZHxwYXJhbXxzb3VyY2V8dHJhY2t8d2JyL2k7XG4gIFxuY29uc3Qgdm9pZEVsZW1lbnRzID0gW1xuICAgIFwiYXJlYVwiLFxuICAgIFwiYmFzZVwiLFxuICAgIFwiYmFzZWZvbnRcIixcbiAgICBcImJnc291bmRcIixcbiAgICBcImJyXCIsXG4gICAgXCJjb2xcIixcbiAgICBcImNvbW1hbmRcIixcbiAgICBcImVtYmVkXCIsXG4gICAgXCJmcmFtZVwiLFxuICAgIFwiaHJcIixcbiAgICBcImltYWdlXCIsXG4gICAgXCJpbWdcIixcbiAgICBcImlucHV0XCIsXG4gICAgXCJpc2luZGV4XCIsXG4gICAgXCJrZXlnZW5cIixcbiAgICBcImxpbmtcIixcbiAgICBcIm1lbnVpdGVtXCIsXG4gICAgXCJtZXRhXCIsXG4gICAgXCJuZXh0aWRcIixcbiAgICBcInBhcmFtXCIsXG4gICAgXCJzb3VyY2VcIixcbiAgICBcInRyYWNrXCIsXG4gICAgXCJ3YnJcIlxuXSIsInRlc3RzIjpbeyJuYW1lIjoiRmluZCBpdGVtIDEwMCIsImNvZGUiOiJ2b2lkRWxlbWVudHMuaW5jbHVkZXMoJ2xpbmsnKSIsInJ1bnMiOls4MTIwMDAsMjAzMzAwMCwxMDc5MDAwLDEyMDEwMDAsMTU4OTAwMCwzMzEwMDAwLDMzNjAwMDAsMjgzNzAwMCwxOTIxMDAwLDExMjQwMDAsNzA0MDAwLDc3MDAwLDEyMjYwMDAsOTEwMDAsNTUwMDAsMTYzNjAwMCwxNDAwMDAwLDEyNDAwMCwxMDM1MDAwLDMxMDkwMDAsMjU5ODAwMCw4NjIwMDAsMTYzNjAwMCwxMzExMDAwLDIwMjAwMDAsMTc0MjAwMCwzMjAwMDAsMTE5NjAwMCwxMDA3MDAwLDE2ODgwMDAsMTQxMDAwLDE2MTYwMDAsNjAwMCwxOTU4MDAwLDc3ODAwMCw3MDYwMDAsMjc4MDAwLDE1MTcwMDAsODI2MDAwLDE3NjAwMDAsMTAwMCwxNzQwMDAsMjExMDAwLDE2MzkwMDAsMjIwMzAwMCw5NTcwMDAsMjI5MDAwLDI0MTAwMDAsODQ2MDAwLDYzMTAwMCw5MjAwMDAsMTA3MDAwLDM0NTQwMDAsMjQ1MzAwMCw2OTIwMDAsMjA1NjAwMCwyNzkzMDAwLDE1MzYwMDAsMTcwNzAwMCwxNzMwMDAsMTEwMDAwMCwxMjMxMDAwLDIwMDAwLDc3OTAwMCwyMDEwMDAsMTkzNjAwMCwyMTY0MDAwLDczMTAwMCw2NDIwMDAsMjAwMDAsMzIzNjAwMCwzNDUwMDAsMjM1MTAwMCwyMDM1MDAwLDI1MjAwMCwxMTQzMDAwLDc4OTAwMCw1MTYwMDAsMjYzMjAwMCwxMTY3MDAwLDk5NzAwMCwyMjEwMDAsNzQ2MDAwLDE3NDQwMDAsMTkwMDAsNjQyMDAwLDE5MDAwLDM4OTAwMCwxMDU3MDAwLDI3MjAwMCwyNzIwMDAwLDE4MTMwMDAsMTc1OTAwMCw0MjQwMDAsMTQ1NDAwMCw0OTUwMDAsMTI1MzAwMCw2MTAwMDAsMzIwMDAwLDI0OTQwMDBdLCJvcHMiOjEyMDYxOTB9LHsibmFtZSI6IkZpbmQgaXRlbSAyMDAiLCJjb2RlIjoiSVNfVk9JRF9FTEVNRU5ULnRlc3QoJ2xpbmsnKSIsInJ1bnMiOlsxMDAwLDIwMDAsMTAwMCwyMDkxMDAwLDQ4MzAwMCwxMDAwLDExMDcwMDAsMTI4MDAwMCw2OTkwMDAsMTAwMCwxMDAwLDYyNjAwMCw1NDAwMCwyMDAwLDEwMDAsNzQxMDAwLDE3MTAwMCwyMDAwLDE4MTkwMDAsMTAwMCw3MzAwMDAsMjAwMCwxMDAwLDYwNzAwMCw2NjcwMDAsMTAwMCw3NjEwMDAsMjEzMDAwLDIwMDAsMTAwMCw1MDEwMDAsMjAwMCw2NDAwMDAsNTEwMDAwLDg5NjAwMCw1NjMwMDAsNTg3MDAwLDEwMDAsMjAwMCwyMDAwLDEwMDAsMTAwMCwyNjkwMDAsMTQ1MDAwLDQ5MjAwMCwyMDAwMCwxNTE2MDAwLDEwMDAsMTIyMDAwLDYxNzAwMCwyMDAwLDEwMDAsMjEyOTAwMCw3NjMwMDAsMjAwMCwxNDE4MDAwLDMzODAwMCw0ODYwMDAsMTAwMCwzNjMwMDAsNjkxMDAwLDEyODAwMCwxMDAwLDIwMDAsMjAwMCwxMzUyMDAwLDIwMDAsNDI2MDAwLDI3ODAwMCwxMDgwMDAwLDU4MDAwLDEwMDAsMTYyNjAwMCwxMDAwLDEwMDAsMTE3NTAwMCwxOTAwMDAsMjAwMCw1ODYwMDAsMjAwMCwyMDAwLDIwMDAsMTkwMDAsMTAwMCwyMDAwLDIwMDAsMTAwMCwxOTAwMCwyMDAwLDIzMzAwMCwxMDAwLDI3ODAwMCwyMDAwLDEwMDAsMTc2MDAwLDEwMDAsMTAwMCwyMDAwLDEwMDAsMjAwMF0sIm9wcyI6MzI4MTIwfV0sInVwZGF0ZWQiOiIyMDIzLTAyLTE3VDA0OjEzOjEwLjc4MloifQ%3D%3D

SukkaW: But [`Set.prorotype.has` should be faster than `Array.prototype.includes`].

aidenybai:
@SukkaW @QuiiBz One problem I'm facing is high creation costs compared to respective libraries. Any red flags in the current PR code that may lead to higher DOM creation cost?

image

packages/block/template.ts
aidenybai: Implemented with Set

packages/block/block.ts
```diff
@@ -0,0 +1,216 @@
...
+        get(_, prop: string): Hole {
+          if (prop === cachedProp) return cachedHole;
+          const hole = new Hole(prop);
+          cachedProp = prop;
+          cachedHole = hole;
+          return hole;
+        },
+      },
+    ),
+  );
```
SukkaW, line 38: @aidenybai IMHO, this could be an implementation error here.

Here I have an example:

```tsx
const block = createBlock(({ text, className }) => <div className={className}>{text}</div>);

block({ text: 'Million.js is the Best!', className: 'text-white' });
block({ text: 'Million.js is the Fastest!', className: 'text-black' });
```

Let's break it up:

```
- text: 'Million.js is the Best!'

cachedProp -> text
cachedHole -> new Hole<'text'>

- className: 'text-white'

cachedProp<'text'> !== 'className'
cachedHole -> new Hole<'className'>

- text: 'Million.js is the Fastest!'

cachedProp<'className'> !== 'text'
cachedHole -> new Hole<'text'>
```

And now, the `Hole` instance is being created constantly, causing a huge performance issue.

SukkaW: My original idea is that we could hold a cache for every block:

```ts
export const createBlock = <P = {}>(fn: ((prop: P) => VElement)) => {
  const cache = new Map<keyof P, P[keyof P]>();
  const vnode = fn(new Proxy(
    {},
    {
      // A universal getter will return a Hole instance if props[any] is accessed
      // Allows code to identify holes in virtual nodes ("digs" them out)
      get(_, prop: string): Hole {
        if (cache.has(prop)) return cache.get(prop);
        const hole = new Hole(prop);
        cache.set(prop, hole);
        return hole;
      },
    },
  ));
}
```

SukkaW: My original idea about `lastProp` was that: the shape of `prop` could have changed, so we might want to re-create every hole instance when the prop shape changes.

But it seems that I was wrong, the shape of the prop is not going to change. So let's throw away the `lastProp` here.

aidenybai: Makes sense, seems like a very reasonable approach + thanks for the helpful examples! Pushing a change now.

aidenybai: image

SukkaW: Have you set up a benchmark suite for new Block API yet? I'd like to try it locally to see if I can spot anything.

packages/block/dom.ts
```diff
@@ -0,0 +1,178 @@
...
+    !Reflect.has(el.style, name) &&
+    !(el instanceof SVGElement) &&
+    name !== 'href' &&
+    name !== 'list' &&
+    name !== 'form' &&
+    // cast to `0` instead
+    name !== 'tabIndex' &&
+    name !== 'download' &&
+    name in el
+  ) {
```
QuiiBz, line 162: This whole condition can be faster by using `!Set#has(name)` and moving `name in el` as the very beginning (as a fast path)

SukkaW: And another issue. This condition uses both `!Reflect.has(el.style, name)` and `name in el`. Ideally, this should be consistent (choose only one of them).

aidenybai: Fixed, list of props hoisted to a set, and chose `name in el` over Reflect

packages/block/block.ts
aidenybai: Old benchmarks will be deprecated in favor of JS Framework Benchmark

aidenybai:
image
Here are the latest benchmarks

packages/block/block.ts
```diff
@@ -0,0 +1,212 @@
...
+    key?: string,
+  ) {
+    super();
+    this.root = root;
+    this.props = props;
+    this.edits = edits;
+    // Cache for getCurrentElement()
+    this.cache = new Map<number, HTMLElement>();
+    this.key = key;
+  }
```
SukkaW, line 81:
```suggestion
export class Block extends AbstractBlock {
  root: HTMLElement;
  edits: Edit[];
  // Cache for getCurrentElement()
  cache = new Map<number, HTMLElement>();

  constructor(
    root: HTMLElement,
    edits: Edit[],
    props?: Props | null,
    key?: string,
  ) {
    super();
    this.root = root;
    this.props = props;
    this.edits = edits;
    this.key = key;
  }
```

We can move the `cache`'s initialization out of the constructor, to make the code cleaner (It means that `cache` can be inited without arguments of the constructor).

packages/block/block.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  // Handles case for positioning text nodes. When text nodes are
+  // put into a template, they can be merged. For example,
+  // ["hello", "world"] becomes "helloworld" in the DOM.
+  // Inserts text nodes into the DOM at the correct position.
+  for (let i = 0, j = edits.length; i < j; ++i) {
+    const current = edits[i]!;
+    if (!current.inits.length) continue;
+    const el = getCurrentElement(current, root);
+    for (let k = 0, l = current.inits.length; k < l; ++k) {
```
SukkaW, line 54:
```suggestion
    const l = current.inits.length;
    if (!l) continue;
    const el = getCurrentElement(current, root);
    for (let k = 0; k < l; ++k) {
```

Let's hoist the cache of `current.inits.length`, so that we could avoid a potential extra `.length` access.

packages/block/block.ts
```diff
@@ -0,0 +1,212 @@
...
+  slot?: number, // edit index
+): HTMLElement => {
+  if (!current.path.length) return root;
+  if (cache && slot !== undefined && mapHas$.call(cache, slot)) {
+    return mapGet$.call(cache, slot)!;
+  }
+  // path is an array of indices to traverse the DOM tree
+  // For example, [0, 1, 2] becomes root.childNodes[0].childNodes[1].childNodes[2]
+  // We use path because we don't have the actual DOM nodes until mount()
+  for (let k = 0, l = current.path.length; k < l; ++k) {
```
SukkaW, line 202:
```suggestion
  const l = current.path.length;
  if (!l) return root;
  if (cache && slot !== undefined && mapHas$.call(cache, slot)) {
    return mapGet$.call(cache, slot)!;
  }
  // path is an array of indices to traverse the DOM tree
  // For example, [0, 1, 2] becomes root.childNodes[0].childNodes[1].childNodes[2]
  // We use path because we don't have the actual DOM nodes until mount()
  for (let k = 0; k < l; ++k) {
```

packages/block/fragment.ts
```diff
@@ -0,0 +1,159 @@
...
+    this._parent = parent;
+    return parent;
+  }
+  remove() {
+    const parent = this.parent;
+    if (parent) {
+      setTextContent$.call(parent, '');
+    } else {
+      for (let i = 0, j = this.children.length; i < j; ++i) {
+        this.children[i]!.remove();
```
SukkaW, line 148: Should use the cached prototype `remove$` here.

packages/block/dom.ts
```diff
@@ -0,0 +1,172 @@
...
+      el.style.cssText = value;
+    } else if (name.startsWith('-')) {
+      el.style.setProperty(name, String(value));
+    } else if (isValueNully) {
+      el.style[name] = '';
+    } else if (typeof value !== 'number' || IS_NON_DIMENSIONAL.test(name)) {
+      el.style[name] = value;
+    } else {
+      el.style[name] = `${String(value)}px`;
+    }
```
SukkaW, line 141: IMHO, the number prop should be more common than CSS variable prop. Let's lift `typeof value === 'number'` above.

aidenybai:
image

This is with new changes and improvement in benchmark implementation: https://github.com/aidenybai/js-framework-benchmark

aidenybai:
image
Also, seems like new Million has high memory usage. Any way we can get the GC to work?

QuiiBz:
Most of the time is spent inside `patch()`, and the memory increases by +2MB (running "create 1k rows"):

Screenshot 2023-02-19 at 08 52 46

I'm not sure to understand **why** we would run `patch()` after `render()`, because calling `render()` already does what is expected. I've tried commenting this line and is still works as expected, could you explain to me why this is needed?

Screenshot 2023-02-19 at 08 52 30

packages/block/types.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export const enum EditType {
+  Attribute,
+  Block,
+  Child,
+  Event,
+}
```
SukkaW, line 35: `unbuild` uses `esbuild` under the hood, which doesn't support cross-file inlining. So here is what `unbuild` outputs:

```js
var EditType = /* @__PURE__ */ ((EditType2) => {
  EditType2[EditType2["Attribute"] = 0] = "Attribute";
  EditType2[EditType2["Block"] = 1] = "Block";
  EditType2[EditType2["Child"] = 2] = "Child";
  EditType2[EditType2["Event"] = 3] = "Event";
  return EditType2;
})(EditType || {});
```

You can verify it by running `pnpm run build` and checking the `dist` folder.

This could drastically reduce the performance.

Before `esbuild` supports cross-file inlining, we should avoid using `const enum`. I propose this instead:

```js
export const EditTypeAttribute = 1
export const EditTypeBlock = 2
export const EditTypeChild = 3
export const EditTypeEvent = 4
```

aidenybai:
> Most of the time is spent inside `patch()`, and the memory increases by +2MB (running "create 1k rows"):
> 
> Screenshot 2023-02-19 at 08 52 46

I wonder how this can be fixed? I assume it's an issue with running the edits. Would it help if the actual DOM changes are deferred until the edits are computed?

packages/block/types.ts
aidenybai: Got it, changing now.

QuiiBz:
> I wonder how this can be fixed? I assume it's an issue with running the edits. Would it help if the actual DOM changes are deferred until the edits are computed?

The most consuming methods are `getCurrentElement()`, `insertText()` and `createEventListener()`. I'm not sure if these can be optimized by a lot, as the culprit seems to be the DOM methods:

Screenshot 2023-02-19 at 10 51 40

packages/block/dom.ts
```diff
@@ -0,0 +1,177 @@
...
+};
+
+export const insertText = (
+  el: HTMLElement,
+  value: any,
+  index: number,
+): Text => {
+  const node = document.createTextNode(value);
+  const childNodes = childNodes$.call(el);
+  if (hasChildNodes$.call(el) && index < childNodes.length) {
```
QuiiBz, line 116: Fast path that can avoid a `hasChildNodes()` call:
```suggestion
  if (index < childNodes.length && hasChildNodes$.call(el)) {
```

packages/block/template.ts
```diff
@@ -0,0 +1,147 @@
...
+        value: undefined,
+        patch: undefined,
+        block: undefined,
+      });
+      continue;
+    }
+
+    if (child instanceof AbstractBlock) {
+      current.edits.push({
+        type: EditType.Block,
```
SukkaW, line 112:
```suggestion
        type: 'child',
        hole: child.key,
        index: i,
        name: undefined,
        listener: undefined,
        value: undefined,
        patch: undefined,
        block: undefined,
      });
      continue;
    }

    if (child instanceof AbstractBlock) {
      current.edits.push({
        type: 'block',
```

packages/block/dom.ts
SukkaW: Also:

```
  const childNodes = childNodes$.call(el);
  if (hasChildNodes$.call(el) && index < childNodes.length) {
```

`childNodes` is accessed even before `hasChildNodes` is called, thus it is possible to eliminate that extra `childNodes` access.

packages/block/dom.ts
```diff
@@ -0,0 +1,177 @@
...
+  value?: string | boolean | null,
+): void => {
+  const isValueNully = value === undefined || value === null;
+  value = isValueNully ? '' : value;
+  if (name === 'style') {
+    if (typeof value !== 'number' || IS_NON_DIMENSIONAL.test(name)) {
+      el.style[name] = value;
+    } else if (typeof value === 'string') {
+      el.style.cssText = value;
+    } else if (name.startsWith('-')) {
```
SukkaW, line 140:
```suggestion
    } else if (name[0] === '-') {
```

packages/block/dom.ts
```diff
@@ -113,38 +107,62 @@ export const insertText = (
 ): Text => {
   const node = document.createTextNode(value);
   const childNodes = childNodes$.call(el);
+<<<<<<< HEAD
+  insertBefore$.call(el, node, childNodes[index]);
+=======
```
SukkaW:
```suggestion
<<<<<<< HEAD
  insertBefore$.call(el, node, childNodes[index]);
=======
```

Wait what?

aidenybai: I think that might be part of an older commit, I can't seem to find it in the Files Changed tab

aidenybai:
@SukkaW @QuiiBz After a couple days of research, I have concluded the memory leak / problems come down to 3 major problems (that I am unsure of how to solve):

1. Row cache in JS Framework Benchmark implementation ([see code](https://github.com/aidenybai/js-framework-benchmark/blob/b0c2f4f1e5803cfd8b097b046f1761b02c3e1bb9/frameworks/keyed/million/src/main.js#L216))

Despite using a new cache object for each render, the old cached blocks seem to linger around somehow, causing a memory leak. Deleting the cache fixes the memory leak, but the cache is really good for creation benchmarks (delta of 10ms)

2. Fragment children not being GC'd ([see code](https://github.com/aidenybai/million/blob/8681a2c31d3eb20b09e4ed84e9904843c61068d2/packages/block/fragment.ts#L28))

When patching two fragment blocks, the old children is saved somehow, causing a huge increase in memory usage. For some reason, blocks refuse to be GC'd.

3. Any caches hold up a ton of memory and refuse to be GC'd ([1](https://github.com/aidenybai/million/blob/8681a2c31d3eb20b09e4ed84e9904843c61068d2/packages/block/block.ts#L23), [2](https://github.com/aidenybai/million/blob/8681a2c31d3eb20b09e4ed84e9904843c61068d2/packages/block/block.ts#L73))

I think the common theme here is that blocks are not getting GC'd. Any suggestions / ideas would be appreciated.

packages/block/block.ts
```diff
@@ -0,0 +1,242 @@
...
+  mapSet$,
+  setStyleAttribute,
+  setSvgAttribute,
+} from './dom';
+import { renderToTemplate } from './template';
+import { AbstractBlock, Hole } from './types';
+import type { Edit, EditChild, Props, VElement } from './types';
+
+export const createBlock = (fn: (props?: Props) => VElement) => {
+  const cache = new Map<string, Hole>();
```
SukkaW, line 23: @aidenybai Most of the time, the `createBlock` will be called at the top level of a module:

```tsx
const row = createBlock(({ value }) => {
  return <li>{value}</li>;
});
```

Thus the block will never be GC-ed, which also means the `cache` will never be released as well.

packages/block/types.ts
```diff
@@ -0,0 +1,93 @@
+export type { VNode, VElement } from '../million/types';
+export type Props = Record<string, any>;
+
+export class Hole {
+  key: string;
+  constructor(key: string) {
+    this.key = key;
+  }
+}
```
SukkaW, line 9: Do we really need a class here? This only holds a string key.

aidenybai: switched to object!

QuiiBz:
I wanted to check the perf, but looks like a regression has been introduced within the last commits:

Screenshot 2023-02-25 at 15 53 00

Screenshot 2023-02-25 at 15 54 07

packages/block/block.ts
```diff
@@ -0,0 +1,248 @@
...
+    new Proxy(
+      {},
+      {
+        // A universal getter will return a Hole instance if props[any] is accessed
+        // Allows code to identify holes in virtual nodes ("digs" them out)
+        get(_, key: string): Hole {
+          if (mapHas$.call(cache, key)) return mapGet$.call(cache, key);
+          const hole = { __key: key };
+          mapSet$.call(cache, key, hole);
+          return hole;
```
SukkaW, line 36: Also, if `Hole` is not going to be complicated, do we really need a cache here?

aidenybai:
@QuiiBz pushed a fix, try again

packages/block/block.ts
aidenybai: probably not. good point here

packages/block/template.ts
```diff
@@ -0,0 +1,153 @@
...
+        value: undefined,
+        index: undefined,
+        patch: undefined,
+        block: undefined,
+      });
+
+      continue;
+    }
+
+    if ('__key' in value) {
```
QuiiBz, line 70: Still having the same bug on this line. Checking if the value is an object before does solve the issue:

```suggestion
    if (typeof value === 'object' && '__key' in value) {
```

Screenshot 2023-02-25 at 21 38 04

aidenybai:
**Benchmark setup for reference:**

1. Clone https://github.com/aidenybai/js-framework-benchmark
2. Setup binaries for benchmarking: https://github.com/aidenybai/js-framework-benchmark#1-new-run-pre-built-binaries-for-all-frameworks
3. Make sure you have three terminal instances running
- 1. Start dev server: `npm run start`
- 2. Build the bundle for Million:
```
cd frameworks/keyed/million
npm i
npm build-dev
```
Make sure you modify the path to your `packages/block/index.ts` here: https://github.com/aidenybai/js-framework-benchmark/blob/31f620323c8e9e306937fcc6f1aca7c17f3c3283/frameworks/keyed/million/src/main.js#L5
- 3. Run benchmark:`npm run bench -- --framework keyed/million --benchmark 01_ 02_ 03_ 04_ 05_ 06_ 07_ 08_ 09_ && npm run results`

4. Then open the `./webdriver-ts-results/table.html` file

packages/block/template.ts
```diff
@@ -0,0 +1,153 @@
...
+        value: undefined,
+        index: undefined,
+        patch: undefined,
+        block: undefined,
+      });
+
+      continue;
+    }
+
+    if (typeof value === 'object' && '__key' in value) {
```
SukkaW, line 70: `typeof value === 'object'` is not safe. `null` also has a type of `object`.

packages/block/block.ts
```diff
@@ -0,0 +1,244 @@
...
+  for (let i = 0; i < pathLength; ++i) {
+    const siblings = current.path[i]!;
+    // https://www.measurethat.net/Benchmarks/Show/15652/0/childnodes-vs-children-vs-firstchildnextsibling-vs-firs
+    root = firstChild$.call(root);
+    if (!siblings) continue;
+    for (let j = 0; j < siblings; ++j) {
+      root = nextSibling$.call(root) as HTMLElement;
+    }
+  }
+  if (cache && slot !== undefined) mapSet$.call(cache, slot, root);
```
SukkaW, line 226:
```suggestion
  const useCache = cache && slot !== undefined;
  if (useCache && mapHas$.call(cache, slot)) {
    return mapGet$.call(cache, slot)!;
  }
  // path is an array of indices to traverse the DOM tree
  // For example, [0, 1, 2] becomes:
  // root.firstChild.firstChild.nextSibling.firstChild.nextSibling.nextSibling
  // We use path because we don't have the actual DOM nodes until mount()
  for (let i = 0; i < pathLength; ++i) {
    const siblings = current.path[i]!;
    // https://www.measurethat.net/Benchmarks/Show/15652/0/childnodes-vs-children-vs-firstchildnextsibling-vs-firs
    root = firstChild$.call(root);
    if (!siblings) continue;
    for (let j = 0; j < siblings; ++j) {
      root = nextSibling$.call(root) as HTMLElement;
    }
  }
  if (useCache) mapSet$.call(cache, slot, root);
```

aidenybai: pushed in latest commit!

aidenybai:
in the latest commit you can officially use million inside react:

```js
import { useState } from 'react';
import { createRoot } from 'react-dom/client';
import { optimize } from '/Users/aidenybai/Projects/aidenybai/million/packages/react'; // replace this with your path

function Display({ foo }) {
  const vnode = (
    <div>
      <div>Hello: {foo}</div>
      <div>
        {Array(50)
          .fill(0)
          .map((_, i) => (
            <div></div>
          ))}
      </div>
    </div>
  );
  return vnode;
}

const OptimizedDisplay = optimize(Display);

const Counter = () => {
  const [count, setCount] = useState(0);
  const [optimized, setOptimized] = useState(false);
  return (
    <div>
      <button onClick={() => setOptimized(!optimized)}>
        {optimized ? 'use unoptimized instead' : 'use optimized instead'}{' '}
      </button>
      <div>
        Count: {count}, optimized?{' '}
        <span style={{ background: optimized ? 'green' : 'red' }}>
          {String(optimized)}
        </span>
      </div>
      <button onClick={() => setCount(count + 1)}>Increment</button>
      <div>
        {Array(1000)
          .fill(0)
          .map((_, i) => {
            return optimized ? (
              <OptimizedDisplay foo={count * i} />
            ) : (
              <Display foo={count * i} />
            );
          })}
      </div>
    </div>
  );
};

createRoot(document.getElementById('root')).render(<Counter />);

```

QuiiBz:
Amazing work! I'm wondering if we could somehow remove the extra `div` that is added to the DOM and given a `ref`? That could easily break the CSS:

Screenshot 2023-02-27 at 07 38 23

aidenybai:
@QuiiBz Not sure if this new commit will work for all cases, but seems to work with the demo.

QuiiBz:
I've re-run the js-framework-benchmark, that's fast! Only append and replace are faster in solid:

Screenshot 2023-02-27 at 09 12 57

SukkaW:
> I've re-run the js-framework-benchmark, that's fast! Only append and replace are faster in solid:

What about the memory usage? Has the memory leak issue been solved yet?

packages/react/react.ts
```diff
@@ -36,14 +46,30 @@ export const optimize = (
...
-        : createElement('div', { style: { display: 'none' }, ref: marker }),
+      createElement(Marker, { ref: marker }),
       createElement(Effect, { effect }),
     );
   };
 };
 
+const Marker: FunctionComponent<{ ref: RefObject<HTMLDivElement> }> = memo(
+  forwardRef((_, ref: any) => {
+    useInsertionEffect(() => {
```
aidenybai, line 57: Currently, the implementation first replaces the temporary <div> marker with the actual component, then on unmount it replaces it back in order to respect the React renderer. This is _very_ expensive, is there any way to trick the React renderer to pretend that this node doesn't exist?

QuiiBz:
> Cool! What about the memory usage? Has the memory leak issue been solved yet?

The memory leak seems to be gone, you can see the heap in blue getting garbage collected (running 5x times Create 1k rows + clear):

Screenshot 2023-02-27 at 09 43 46

packages/react/react.ts
aidenybai: I'll take a look and try to learn what they're doing. If you have any ideas of how this could be implemented (I think our case should be much easier than signals), feel free to suggest

packages/million/template.ts
```diff
@@ -0,0 +1,173 @@
...
+        listener: undefined,
+        value: undefined,
+        index: undefined,
+        patch: undefined,
+        block: undefined,
+      });
+      continue;
+    }
+
+    if (value) props += ` ${name}="${String(value)}"`;
```
SukkaW, line 99: This should guard against `value` being `null` (`typeof null === 'object'` is `true`, but you can't use `in` on `null`).

```suggestion
    if (value) {
      if (typeof value === 'object' && '__key' in value) {
        current.edits.push({
          type:
            name === 'style'
              ? 'style'
              : name.charCodeAt(0) === X_CHAR
              ? 'svg'
              : 'attribute',
          hole: value.__key,
          name,
          listener: undefined,
          value: undefined,
          index: undefined,
          patch: undefined,
          block: undefined,
        });
        continue;
      }
      props += ` ${name}="${String(value)}"`;
    }
```

packages/million/template.ts
```diff
@@ -0,0 +1,173 @@
...
+    if (name === 'htmlFor') name = 'for';
+
+    if (name.startsWith('on')) {
+      const isValueHole = '__key' in value;
+      // Make objects monomorphic
+      current.edits.push({
+        type: 'event',
+        listener: isValueHole ? value.key : value,
+        name,
+        hole: isValueHole ? value.key : undefined,
```
SukkaW, line 70: Is it `value.key` or `value.__key`? I am a bit confused here.

aidenybai: this should be value.__key, nice catch

packages/million/template.ts
aidenybai: good catch!

aidenybai:
I may need some help with tests + docs if anyone wants to get started on that:

For tests, we'll need to cover most of the new Block API under `packages/million`, not too concerned about the React API for now because it's pretty experimental ([example test code](https://github.com/aidenybai/million/blob/main/test/render.test.ts))

We'll also need to rewrite the documentation so that it removes the existing content and adds in the new Block API reference.

packages/million/block.ts
```diff
@@ -0,0 +1,262 @@
...
+    refNode: Node | null = null,
+    inPlace = false,
+  ): HTMLElement {
+    if (this.el) return this.el;
+    // cloneNode(true) uses less memory than recursively creating new nodes
+    const root = cloneNode$.call(this.root, true) as HTMLElement;
+
+    for (let i = 0, j = this.edits.length; i < j; ++i) {
+      const current = this.edits[i]!;
+      const el = getCurrentElement(current, root, this.cache, i);
```
SukkaW, line 103: Continues from https://github.com/aidenybai/million/pull/292#discussion_r1118877216

> `edits` act like a blueprint for an instance of a Block to reference to update based on props.
> 
> [(click on reference)](https://github.com/aidenybai/million/blob/3ebd41e599be86d572bba87f54fa2e88b19e0ddc/packages/million/block.ts#L45) An empty `edits` array is passed into the `renderToTemplate` code. Inside that function, the edits array is mutated ([ref](https://github.com/aidenybai/million/blob/3ebd41e599be86d572bba87f54fa2e88b19e0ddc/packages/million/template.ts#L170)). Later on, the `edits`array will be passed into a new Block instance ([code](https://github.com/aidenybai/million/blob/3ebd41e599be86d572bba87f54fa2e88b19e0ddc/packages/million/block.ts#L68))
> 
> The `edits` will never be changed on mount/patch. However, during mount/patch it is iterated through to apply edits to the Block's element (`this.el`)

Since the `current` here could be considered immutable, we can use `current` as the cache key (`cache.has(current)`) instead of the numeric index. And then we can use `WeakMap` (instead of `Map`) for caching, then the cached HTMLElement could be released if `current` ever get GC-ed.

packages/million/block.ts
```diff
@@ -0,0 +1,262 @@
...
+          setSvgAttribute(el, edit.name, value);
+        }
+      }
+    }
+
+    if (parent) {
+      if (inPlace) {
+        while (root.childNodes.length > 0) {
+          parent.appendChild(root.childNodes[0] as Node);
+        }
```
SukkaW, line 142: Should we cache `childNodes` calls?

```ts
let rootChildNodes;
while ((rootChildNodes = root.childNodes).length > 0) {
  parent.appendChild(rootChildNodes[0] as Node);
}
```

And shouldn't we use cached prototype `childNodes$` here?

Also, I am wondering if `firstChild` is faster:

```ts
let rootFirstChild;
while ((rootFirstChild = root.firstChild)) {
  parent.appendChild(rootFirstChild);
}
```

packages/million/block.ts
aidenybai: Seems like there is not a big difference in memory usage after bench, but change made anyway

packages/million/block.ts
aidenybai: firstChild node access is def the faster way, thanks for the suggestion

packages/million/block.ts
```diff
@@ -0,0 +1,261 @@
...
+  ) => {
+    return new Block(root, edits, props, key ?? props?.key, shouldUpdate);
+  };
+};
+
+export class Block extends AbstractBlock {
+  root: HTMLElement;
+  edits: Edit[];
+  // Cache for getCurrentElement()
+  cache = new WeakMap<Edit, HTMLElement>();
```
SukkaW, line 76: I just realize why `cache` is never getting released.

Imagine this case:

```ts
const block = createBlock();

const row1 = block({}); // This initiates a new Block instance, along with a `.cache`
const row2 = block({}); // This also initiates a new Block instance. Now we have 2 Block instances
// ...
const row1000 = block({}); // Now we have 1000 Block instances
```

If those rows never get GC-ed, so do those 1000 Block instances, and so do those 1000 caches.

aidenybai: Yeah, edits should and will never get GC'd because createBlock acts almost like creating a "component." That's why I thought the numeric index may have been better, but I don't think Maps can be GC'd which means the memory usage is roughly the same

SukkaW: > That's why I thought the numeric index may have been better

We might need a benchmark on this.

Since `edits` will never get GC-ed, the `Map` should be preferred over `WeakMAap`, as `WeakMap` is definitely slower.

aidenybai: benchmarks have roughly the same results

aidenybai:
new homepage @SukkaW  @QuiiBz !

https://million-git-refactor-core-aidenybai.vercel.app/

website/components/chart.tsx
```diff
@@ -0,0 +1,85 @@
...
+    } else {
+      defaults.borderColor = '#bdbfc7';
+      defaults.color = '#545864';
+      defaults.font.family =
+        'ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace';
+    }
+    setCanRender(true);
+  }, []);
+  return (
+    <div className=" p-4 rounded-lg w-auto">
```
SukkaW, line 66: Is it possible to add a `min-height` or placeholder here?

The page's First Paint doesn't include the chart, then the chart shows up, causing a Layout Shift.

aidenybai: hmm good idea

SukkaW:
> https://million-git-refactor-core-aidenybai.vercel.app/

@aidenybai Also, another idea about chart design:

use vanilla as `1.0` like this:

image

aidenybai:
> > [million-git-refactor-core-aidenybai.vercel.app](https://million-git-refactor-core-aidenybai.vercel.app/)
> 
> @aidenybai Also, another idea about chart design: Use vanilla as `1.0` like this:
> 
> image

Updated, let me know what you think!

QuiiBz:
Small flicker issue on mobile:

https://user-images.githubusercontent.com/43268759/223639843-3b0a3062-a770-4bba-9837-6a363e42e050.MOV

Otherwise it's great!

SukkaW:
> https://million-git-refactor-core-aidenybai.vercel.app/

image

@aidenybai Instead of showing "how much is a framework slower than the vanilla", I'd like to say "how fast is a framework compared to vanilla".

So, let's set vanilla as `100%`, then Million `104%`, Solid.js `105%`, and Inferno `109%`, etc. What would you think?

aidenybai:
todo: add react and nextjs integration guide

SukkaW (changes requested):

website/components/chart.tsx
```diff
@@ -0,0 +1,86 @@
...
+  return (
+    <div className="p-4 rounded-lg w-auto min-h-[270px]">
+      {darkMode !== null && (
+        <Bar
+          options={options}
+          data={{
+            labels: data.map((row) => row.framework),
+            datasets: [
+              {
+                label: '% slower than vanilla JavaScript',
```
SukkaW, line 75:
```suggestion
                label: '% of vanilla JavaScript',
```

SukkaW (dismissed):

aidenybai:
@QuiiBz @SukkaW Docs, tests, and code is ready! We are ready to ship if unless there are any issues

SukkaW (commented):
image

@aidenybai Maybe we should make the hero page take fullscreen for now (we can always add more content later)

packages/compiler/babel.ts
```diff
@@ -0,0 +1,64 @@
+import * as t from '@babel/types';
+import type * as babel from '@babel/core';
```
SukkaW, line 2: Just a note, we can create an swc plugin for this in the future.

aidenybai: yes, for sure. next optimization for the compiler should be flattening blocks, not just props diff gen

PR #296, opened 2023-03-15
feat: optimizing compiler

aidenybai:

**Status**

  - [x] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)


packages/compiler/index.ts
```diff
@@ -1,24 +1,29 @@
...
+    enforce: 'pre',
+    name: 'million',
     transformInclude(id: string) {
       return /\.[jt]sx$/.test(id);
     },
     async transform(code: string) {
-      const result = await babel.transformAsync(code, { plugins: [plugin] });
-      return result.code;
+      const result = await transformAsync(code, {
+        plugins: ['@babel/plugin-syntax-jsx', plugin],
```
SukkaW, line 17: We should include typescript parse support here by default as well.

SukkaW, line 17:
```suggestion
        parseOpts: {
      plugins: ['typescript', 'jsx']
    },
```

Do we only need JSX / TS parsing support?

aidenybai: good idea, does babel support ts on its own? Also, when I tried adding parseOpts the types told me it wasn't a valid field

PR #300, opened 2023-03-25
feat: linked list exploration

aidenybai:
thoughts? @SukkaW @QuiiBz performance benchmarks are pretty inconsistent for me, not sure whether this is faster than v2.1.0 or not.

Linked list traversal seems to be ~2x faster than array traversal

### DIFF
--- a/packages/million/dom.ts
+++ b/packages/million/dom.ts
@@ -53,7 +53,7 @@ export const set$ = Set.prototype;
 export const setAdd$ = set$.add;
 export const setHas$ = set$.has;
 
-document[EVENTS_REGISTRY] = new Set();
+document[EVENTS_REGISTRY] = {};
 
 let listenerPointer = 0;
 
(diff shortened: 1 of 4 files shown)


cytronicoder:
Oooh, interesting - wouldn't storing elements contiguously in memory be faster than pointers?

aidenybai:
Haven't benchmarked that, but that's a good point

packages/million/dom.ts
```diff
@@ -86,7 +86,7 @@ export const createEventListener = (
         el = el.parentNode;
       }
     });
-    document[EVENTS_REGISTRY].add(event);
```
SukkaW, line 89: AFAIK, the polymorphic object is slow (per hidden class).

Do we have any benchmark for this change?

packages/million/types.ts
```diff
@@ -34,113 +34,129 @@ export abstract class AbstractBlock {
...
-  PATCH,
-  BLOCK,
+export const enum Flags {
+  CHILD = 1,
+  EVENT = 2,
+  ATTRIBUTE = 4,
+  STYLE_ATTRIBUTE = 8,
+  SVG_ATTRIBUTE = 16,
+  BLOCK = 32,
 }
```
SukkaW, line 44: Does esbuild support cross-file inline const enum now? If not, using `export const enum` will increase the bundle size and decrease the performance.

aidenybai:
interesting stuff, but slower than normal object array

PR #301, opened 2023-03-30
feat: object-based edits

aidenybai:
Here is a somewhat different approach that uses iteration but tries to monomorphically use objects instead of arrays

### DIFF
--- a/packages/compiler/visitor.ts
+++ b/packages/compiler/visitor.ts
@@ -1,9 +1,9 @@
 import { addNamed } from '@babel/helper-module-imports';
 import * as t from '@babel/types';
-import { renderToTemplate, renderToString } from './render';
+import { renderToString, renderToTemplate } from './render';
 import { chainOrLogic } from './utils';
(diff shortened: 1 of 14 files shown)


packages/million/dom.ts
```diff
@@ -158,7 +147,7 @@ export const setAttribute = (
     el[name] !== undefined &&
     el[name] !== null &&
     !(el instanceof SVGElement) &&
-    setHas$.call(NON_PROPS, name)
+    NON_PROPS.has(name)
```
SukkaW, line 150: We should keep the prototype cache here. Prototype cache can skip the prototype chain lookup.

SukkaW: **Chrome**

image

**Firefox**

image

aidenybai: good catch!

packages/compiler/visitor.ts
```diff
@@ -209,3 +237,137 @@ export const visitCallExpression = (path: NodePath<t.CallExpression>) => {
     }
   }
 };
+
```
aidenybai, line 240: marker

PR #313, opened 2023-04-14
fix(react): remove ts-expect-error

QuiiBz:

Remove `@ts-expect-error` & `prettier-ignore` from `react` imports. Because we have a root `react.d.ts` definition file due to the `million/react` import, TS picks this file instead of `node_modules/react`

I found this trick of using `paths` inside the tsconfig, not 100% sure that it works.

**Status**


aidenybai:
LGTM, never knew about `paths` before!

PR #331, opened 2023-05-01
Fix typo in virtual-dom.mdx

eniodev:
A small update in the docs:
Line 85: 2. Dirty Checking...
"is diffed the to determine what has changed" should be "is diffed to determine what has changed"


aidenybai:
nice catch :) LGTM

PR #341, opened 2023-05-07
Ignore null, undefined, false in block rendering

pyjun01:

**Status**

```null```, ```undefined```, ```false``` are ignored in ```renderToTemplate```, but not in block mount and patch
https://github.com/aidenybai/million/blob/main/packages/million/template.ts#L28
https://github.com/aidenybai/million/blob/main/packages/million/template.ts#L138

### DIFF
--- a/packages/million/block.ts
+++ b/packages/million/block.ts
@@ -146,6 +146,11 @@ export class Block extends AbstractBlock {
             insertBefore$.call(el, parent, childNodes[edit.i!]);
             continue;
           }
+
+          if (value === null || value === undefined || value === false) {
+            continue
+          }
+
           // insertText() on mount, setText() on patch
           insertText(el, String(value), edit.i!);
         } else if (edit.t & EventFlag) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by aidenybai** [comment]
@pyjun01 can you run a prettier job on this code?

**Comment by pyjun01** [comment]
> @pyjun01 can you run a prettier job on this code?

applied :)

File: packages/million/block.ts [hunk 1]
```diff
@@ -146,6 +146,11 @@ export class Block extends AbstractBlock {
             insertBefore$.call(el, parent, childNodes[edit.i!]);
             continue;
           }
+
+          if (value === null || value === undefined || value === false) {
```
SukkaW, line 150:
```suggestion
          if (value == null || value === false) {
```

This can be simplified.

packages/million/block.ts
```diff
@@ -227,6 +232,15 @@ export class Block extends AbstractBlock {
           if (typeof oldValue === 'function') {
             continue;
           }
+
+          if (
+            newValue === null ||
+            newValue === undefined ||
+            newValue === false
```
SukkaW, line 239:
```suggestion
            newValue == null ||
            newValue === false
```

packages/million/block.ts
pyjun01: Use of ```==``` is against ESLint rule (```eqeqeq``` in ```@vercel/style-guide/eslint/node```)

tobySolutions (approved):
LGTM! Good job @SukkaW!

pyjun01:
Hi @aidenybai! this pr was closed after being approved. is there a problem?

PR #349, opened 2023-05-11
perf: childAt helper and cache

aidenybai:

This PR introduces two performance improvements:

**1. Improve node traversal within edits.**

Currently, we use efficient node traversal (via `firstChild` and `nextSibling`) for paths, but we still use `childNodes` in some places. This PR replaces `childNodes` with a `childAt` helper function, which internally uses the efficient node traversal method.

**2. Cache text nodes once traversed**

Within child edits, we traverse to a certain index and change the text value, for both mount and patch. With this change, traversal is only done at mount to determine the location of the text node, cached on the current path element, then accessed later during patch. This prevents unnecessary traversal and allows us to directly do pinpoint changes.

**Status**

### DIFF
--- a/packages/compiler/react.ts
+++ b/packages/compiler/react.ts
@@ -41,7 +41,7 @@ export const transformReact =
         );
       }
       const componentBinding = path.scope.getBinding(componentId.name);
-      const component = structuredClone(componentBinding?.path.node);
+      const component = t.cloneNode(componentBinding?.path.node);
 
       if (t.isFunctionDeclaration(component)) {
         handleComponent(
--- a/packages/compiler/react.ts
+++ b/packages/compiler/react.ts
@@ -41,7 +41,7 @@ export const transformReact =
         );
       }
       const componentBinding = path.scope.getBinding(componentId.name);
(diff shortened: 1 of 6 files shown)


QuiiBz (commented):
Seems like a big perf improvement, do you have a before/after benchmark?

aidenybai:
Seems like it is minor for the `js-framework-benchmark`. No change in memory / startup performance.

tobySolutions:
Interesting.

tobySolutions:
Really interesting. 🤔🤔🤔

packages/compiler/react.ts
```diff
@@ -41,7 +41,7 @@ export const transformReact =
         );
       }
       const componentBinding = path.scope.getBinding(componentId.name);
-      const component = { ...componentBinding?.path.node };
+      const component = structuredClone(componentBinding?.path.node);
```
SukkaW, line 44:
```suggestion
      const component = t.cloneNode(componentBinding?.path.node);
```

Prefer babel built-in method over `structuredClone`.

Issue #364, opened 2023-05-28
Framework Switcher (vite-react, next.js, preact, etc)


GuiBibeau:
**Description:**

Implement a framework switcher that allows users to quickly switch between different frameworks like vite-react, next.js, preact etc. The switcher should provide a seamless transition and keep the current state, if possible.

**Acceptance Criteria:**

- [ ] - User should be able to switch between different frameworks through a dropdown menu or similar interface.
- [ ] - The state of the current work should be preserved if possible when switching.
- [ ] - The transition between frameworks should be seamless and intuitive for the user.

GuiBibeau:
I think we can now close this ! Thanks Ethan!

PR #375, opened 2023-05-29
basic framework switcher

ethanniser:

for #364

**Status**

- [ ] Code changes have been tested against prettier, or there are no code changes

### DIFF
--- a/playground/app/page.tsx
+++ b/playground/app/page.tsx
@@ -1,9 +1,9 @@
-import { Editor } from '@/components/editor';
+import ClientEntry from '@/components/client';
 
 const Home: React.FC = () => {
   return (
     <main className="f">
-      <Editor />
+      <ClientEntry />
     </main>
   );
 };
--- a/playground/app/page.tsx
+++ b/playground/app/page.tsx
@@ -1,9 +1,11 @@
-import ClientEntry from '@/components/client';
+'use client';
+
+import { Editor } from '@/components/editor';
 
 const Home: React.FC = () => {
   return (
     <main className="f">
-      <ClientEntry />
+      <Editor />
     </main>
   );
 };
--- a/playground/components/client.tsx
+++ b/playground/components/client.tsx
@@ -1,0 +1,6 @@
+'use client';
+import { Editor } from './editor';
+
+export default function ClientEntry() {
+  return <Editor />;
+}
--- a/playground/components/client.tsx
+++ b/playground/components/client.tsx
@@ -1,6 +1,0 @@
-'use client';
-import { Editor } from './editor';
-
-export default function ClientEntry() {
-  return <Editor />;
-}
--- a/playground/components/editor.tsx
+++ b/playground/components/editor.tsx
@@ -18,10 +18,6 @@ import { useAtomValue, atom } from 'jotai';
 
 export const frameworkAtom = atom<Frameworks>('react');
 
-type Props = {
-  framework?: Frameworks;
-};
-
 const FRAMEWORK_TEMPLATE_MAP: Record<Frameworks, SandpackPredefinedTemplate> = {
   nextjs: 'nextjs',
   react: 'vite-react',
@@ -32,7 +28,7 @@ const FRAMEWORK_FILES_MAP: Record<Frameworks, SandpackFiles> = {
   nextjs: nextjsFiles,
 };
 
-export const Editor: React.FC<Props> = () => {
+export const Editor: React.FC = () => {
   const framework = useAtomValue(frameworkAtom);
 
   const template = FRAMEWORK_TEMPLATE_MAP[framework];
--- a/playground/components/editor.tsx
+++ b/playground/components/editor.tsx
@@ -12,7 +12,6 @@ import { files as nextjsFiles } from '@/configurations/nextjs';
 import { MonacoEditor } from './monaco-editor';
 import { GridResizer } from './gridResize';
 import { useRef, useState } from 'react';
-import cls from 'classnames';
 import { FrameworkSwitcher } from './framework-switcher';
 import { useAtomValue, atom } from 'jotai';
 
--- a/playground/components/editor.tsx
+++ b/playground/components/editor.tsx
@@ -14,6 +14,7 @@ import { MonacoEditor } from './monaco-editor';
 import { GridResizer } from './gridResize';
 import { useRef, useState } from 'react';
 import cls from 'classnames';
+import { FrameworkSwitcher } from './framework-switcher';
 
 type Props = {
   framework?: Frameworks;
@@ -29,7 +30,9 @@ const FRAMEWORK_FILES_MAP: Record<Frameworks, SandpackFiles> = {
   nextjs: nextjsFiles,
 };
 
-export const Editor: React.FC<Props> = ({ framework = 'react' }) => {
+export const Editor: React.FC<Props> = () => {
+  const [framework, setFramework] = useState<Frameworks>('react');
+
   const template = FRAMEWORK_TEMPLATE_MAP[framework];
   const files = FRAMEWORK_FILES_MAP[framework];
   const dependencies = {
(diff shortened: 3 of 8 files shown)


ethanniser:
> The state of the current work should be preserved if possible when switching.

Unclear what the goal here is, its pretty easy to retain the code between switched, but different frameworks use different code so I don't really know what to make of this.

Right now I have the shell restarting on a 1 second timeout, as I found any sooner and the new files didnt load and the previous frameworks startup commands would be run, dont know if this passes the *The transition between frameworks should be seamless and intuitive for the user.* bar but could be improved probably.

ethanniser:
also I dont believe sandpack has a preact template so we might need to make a custom setup for that

playground/components/editor.tsx
```diff
@@ -29,7 +30,9 @@ const FRAMEWORK_FILES_MAP: Record<Frameworks, SandpackFiles> = {
   nextjs: nextjsFiles,
 };
 
-export const Editor: React.FC<Props> = ({ framework = 'react' }) => {
+export const Editor: React.FC<Props> = () => {
+  const [framework, setFramework] = useState<Frameworks>('react');
```
GuiBibeau, line 34: Is this state consumed anywhere?

playground/components/editor.tsx
```diff
@@ -65,17 +68,14 @@ export const Editor: React.FC<Props> = ({ framework = 'react' }) => {
   };
   return (
     <div className="flex flex-col h-100vw">
-      <header
-        style={{ height: 50 }}
-      >
-        header
-      </header>
+      <header style={{ height: 50 }}>header</header>
```
GuiBibeau, line 71: Thanks for running prettier. 🙏

playground/components/framework-switcher.tsx
```diff
@@ -0,0 +1,41 @@
+'use client';
+
+import { Frameworks } from '@/types';
+import { useSandpack } from '@codesandbox/sandpack-react';
+import { Dispatch, SetStateAction } from 'react';
```
GuiBibeau, line 5: 💡 idea: 
```
import type { Frameworks } from '@/types';
import { useSandpack } from '@codesandbox/sandpack-react';
import type { Dispatch, SetStateAction } from 'react';
```

This tells the compiler these imports are not needed during runtime.

playground/components/framework-switcher.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export function FrameworkSwitcher({
+  setFramework,
```
GuiBibeau, line 8: I would not drill this down as props. I would have it come from a context.

playground/components/framework-switcher.tsx
```diff
@@ -0,0 +1,41 @@
...
+}: {
+  setFramework: Dispatch<SetStateAction<Frameworks>>;
+}) {
+  const { dispatch } = useSandpack();
+
+  const handleClick = (framework: Frameworks) => {
+    setFramework(framework);
+    setTimeout(() => {
+      dispatch({ type: 'shell/restart' });
+    }, 1000);
```
GuiBibeau, line 18: Why the setTimeout?

playground/components/framework-switcher.tsx
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  return (
+    <>
+      <div className="flex flex-col items-center justify-center h-12">
+        <div className="flex items-center justify-center">
+          <button
```
GuiBibeau, line 25: I would change two things here:

- This should be a select instead of buttons. This will not scale nicely to more frameworks 
- We should dynamically render those based on a configuration object or something where editing will be easier and dynamic in the future.

GuiBibeau (changes requested):
Great first draft! I like the direction but I would love to request a few changes:

- Not drilling into props and going into a context or something similar 
- move from buttons to a select rendered based on configurations.

Let me know if you want to pair or discuss more!

playground/components/framework-switcher.tsx
ethanniser: > Right now I have the shell restarting on a 1 second timeout, as I found any sooner and the new files didnt load and the previous frameworks startup commands would be run, dont know if this passes the*The transition between frameworks should be seamless and intuitive for the user.* bar but could be improved probably.

playground/components/editor.tsx
tobySolutions: Good job folks!

playground/components/framework-switcher.tsx
tobySolutions: Good job folks!!

playground/components/framework-switcher.tsx
tobySolutions: Nice!

playground/components/framework-switcher.tsx
GuiBibeau: Makes a lot of sense to me!

playground/package.json
```diff
@@ -14,14 +14,18 @@
     "@types/react-dom": "18.2.4",
     "add": "^2.0.6",
     "autoprefixer": "10.4.14",
+    "classnames": "^2.3.2",
```
SukkaW, line 17: Prefer `clsx` (https://npm.im/clsx) over `classnames`.

playground/components/editor.tsx
```diff
@@ -14,10 +13,10 @@ import { MonacoEditor } from './monaco-editor';
 import { GridResizer } from './gridResize';
 import { useRef, useState } from 'react';
 import cls from 'classnames';
+import { FrameworkSwitcher } from './framework-switcher';
+import { useAtomValue, atom } from 'jotai';
 
-type Props = {
-  framework?: Frameworks;
-};
+export const frameworkAtom = atom<Frameworks>('react');
```
SukkaW, line 19: IMHO, it is better to use context + useState pattern over `jotai`. We are only trying to prevent the prop drilling and `jotai` is overkill.

GuiBibeau: We had a parallel discussion in Discord, my bad for not surfacing it in the PR. The Jotai move is mostly as because we expect we will go quickly have more state come up.

100% agree that this can be simply context + useState, I personally don't have a preference over each but I think Jotai will come to make sense in the next 2-3 features.

playground/app/page.tsx
```diff
@@ -1,9 +1,9 @@
-import { Editor } from '@/components/editor';
+import ClientEntry from '@/components/client';
```
GuiBibeau, line 1: Is this to organize client components into a specific area of the code?

playground/components/client.tsx
```diff
@@ -0,0 +1,6 @@
+'use client';
+import { Editor } from './editor';
+
+export default function ClientEntry() {
+  return <Editor />;
```
GuiBibeau, line 5: I'm not sure we need a wrapper component. The intent seems to be to abstract the client component entry point. 

While this might sound like an easier mental load and organizing structure, it's adding a new layer of complexity to exports, imports and naming, making refactor difficult.

A flatter component structure will be much easier to refactor.

GuiBibeau (changes requested):
Great changes! Thanks for pushing this along.

 My last comments are only related to `playground/components/client.tsx`. I would prefer that we keep things simple for now and not add a level of nesting too early.

It would be great to also address @SukkaW changes especially for `clsx` since the bundle size is smaller (if I remember)

ethanniser:
The "client entry" was a remnant of my original implementation with context, I needed a client component that wraps "Editor" to provide the context, and I didn't realize you can make 'page.tsx' a client component. I forgot to remove it after I added Jotai.

But I will clean that up, move to clsx and try to fix the sandpack height styles tonight.
Thank you all for the feedback.

ethanniser:
Was unable to fix the styles, tried all kinds of things but sandpack has some weird styling behavior I can't grasp. Although for now hard coding the height of all the other elements is not a huge deal.

GuiBibeau (approved):
Thanks a lot for the hard work! LGTM 🔥

Issue #382, opened 2023-06-06
bug: Support for TypeScript on components that take in props


destroyer22719:
**Describe the bug**
MillionJS doesn't support components that takes in props in TypeScript,

**To Reproduce**
Below is a minimal code that causes this behaviour

```js
import React from "react";
import { block } from "million";

const myBlock = block(({ value }: { value: string }) => {
  return <div>{value}</div>;
});

```
In which TypeScript complains
```
Argument of type '({ value }: { value: string; }) => JSX.Element' is not assignable to parameter of type '(props?: Props | undefined) => VElement'.
  Types of parameters '__0' and 'props' are incompatible.
    Type 'Props | undefined' is not assignable to type '{ value: string; }'.
      Type 'undefined' is not assignable to type '{ value: string; }'.
```

**Expected behavior**
MillionJS' `block` function should be compatible with props in TypeScript as well, as well as ideally having type support for blocks when used in JSX.

**Proposed Solution**
Employ a similar method used by regular React and that is to specify the return type of `React.FC<Props>`
A solution would look like this:
```ts
import React from "react";
import { block, MillionBlock } from "million";

type Props {
  value: string
}

const myBlock: MillionBlock<Props> = block(({ value }) => {
  return <div>{value + 1}</div>; //in which then typescript would know value is a string, nut a number, thus we can't add
});
```

destroyer22719:
This issue has been fixed with PR #385

Issue #368, opened 2023-05-28
Vite React TypeScript Support


GuiBibeau:
**Description:**

Extend support for Vite React framework with TypeScript in the playground. The application should allow users to write and test Vite React code using TypeScript.

**Acceptance Criteria:**

- [ ] User can select Vite React (TypeScript) from the framework switcher.

- [ ] User can select Vite React (TypeScript) from the framework switcher.G

aidenybai:
this is a comment from dimension!

PR #385, opened 2023-06-07
TypeScript support for props in the `block` function and the `For` component

destroyer22719:

**Status**

- [] I know how to update typings and have done so, or typings don't need updating

This PR addresses issues #382 and partially #368.

```tsx
import React from 'react';
import { block, For } from "million/react";


type Props = {
  name: string;
};

const MyComponent: React.FC<Props> = ({ name }) => {
  return <div>Hello {name}</div>;
};

const MyBlock = block<Props>(MyComponent);

const Parent = () => {
  const names = ['John', 'Jane', 'Joe'];

  return (
    <div>
      {/* Name will now be recognized as a string instead of any type, TypeScript will also know that MyBlock takes in a name prop */}
      <For each={names}>{(name) => <MyBlock name={name} />}</For>
    </div>
  );
};
```


kul-sudo:
Shows ```Type 'BlockProps' does not satisfy the constraint 'MillionProps'.```

PR #440, opened 2023-06-18
Playground: Preset Passed Via Search Param in URL

### LINKED ISSUE #365: Preset Passed Via Search Param in URL
**Description:**

Implement a feature that allows presets to be passed via search parameters in the URL. This will allow users to quickly load specific setups or configurations.

**Acceptance Criteria:**

- [ ] User can load a preset by visiting a URL with the preset specified as a search parameter.
- [ ] The playground should correctly load and apply the preset.

ethanniser:
Closes #365 

> Got to a decent point, required some kinda wild useEffect stuff to keep the url + atom in sync. Jotai does provide a atomWithHash, but [unfortunately doesn't support Next 13 / app router](https://jotai.org/docs/guides/nextjs#in-next-js-13)
> 
> Also found a bug where if you select a new framework while the current one is loading it will show the new files, but use the previous start command.
> 
> would appreciate any feedback on what I did so far. Thanks!

through some git stuff I dont understand the first pr got closed but at least now its synced with the the current state of the playground branch

### DIFF
--- a/playground/components/editor.tsx
+++ b/playground/components/editor.tsx
@@ -6,6 +6,9 @@ import {
   SandpackPreview,
 } from '@codesandbox/sandpack-react';
 import { useEffect, useRef, useState } from 'react';
+import { z } from 'zod';
+import { useRouter, useSearchParams } from 'next/navigation';
+import { useAtom, useAtomValue } from 'jotai';
 import { files as reactViteFiles } from '@/configurations/react-vite';
 import { files as nextjsFiles } from '@/configurations/nextjs';
 import { frameworkAtom } from '@/atoms/framework';
@@ -13,9 +16,6 @@ import { MonacoEditor } from './monaco-editor';
 import { GridResizer } from './grid-resizer';
 import { FrameworkSwitcher } from './framework-switcher';
 import type { Framework } from '@/types';
-import { atom, useAtom, useAtomValue } from 'jotai';
-import { useRouter, useSearchParams } from 'next/navigation';
-import { z } from 'zod';
 
 const zFrameworks = z.enum(['react', 'nextjs']);
 
--- a/playground/components/editor.tsx
+++ b/playground/components/editor.tsx
@@ -6,18 +6,14 @@ import {
   SandpackPreview,
 } from '@codesandbox/sandpack-react';
 import { useEffect, useRef, useState } from 'react';
-import { z } from 'zod';
 import { useRouter, useSearchParams } from 'next/navigation';
 import { useAtom, useAtomValue } from 'jotai';
 import { files as reactViteFiles } from '@/configurations/react-vite';
 import { files as nextjsFiles } from '@/configurations/nextjs';
-import { frameworkAtom } from '@/atoms/framework';
+import { frameworkAtom, zFrameworks, type Framework } from '@/atoms/framework';
 import { MonacoEditor } from './monaco-editor';
 import { GridResizer } from './grid-resizer';
 import { FrameworkSwitcher } from './framework-switcher';
-import type { Framework } from '@/types';
-
-const zFrameworks = z.enum(['react', 'nextjs']);
 
 export const useFrameworkSyncUrl = () => {
   const router = useRouter();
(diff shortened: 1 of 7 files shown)


playground/components/framework-switcher.tsx
```diff
@@ -15,7 +15,7 @@ const frameworks: FrameworkConfig[] = [
 
 export function FrameworkSwitcher() {
   const { dispatch } = useSandpack();
-  const setFramework = useSetAtom(frameworkAtom);
+  const [framework, setFramework] = useAtom(frameworkAtom);
 
   const handleSelect = (event: React.ChangeEvent<HTMLSelectElement>) => {
```
brunoeduardodev, line 20: in general, should we call it `framework` or `preset`? like, will be a scenario where we want to have a `next` and `next-ts` presets? in this case it might make more sense calling it PresetSwitcher, and use `preset` on the searchParams instead of `framework`

playground/components/editor.tsx
```diff
@@ -15,6 +17,38 @@ import { GridResizer } from './grid-resizer';
 import { FrameworkSwitcher } from './framework-switcher';
 import type { Framework } from '@/types';
 
+const zFrameworks = z.enum(['react', 'nextjs']);
+
```
brunoeduardodev, line 21: I think we could even move this to `@/atoms/framework` and infer the `Framework` type from the zod schema, this way we don't need to update multiple places every time we add a new preset/framework, any thoughts?

playground/components/framework-switcher.tsx
brunoeduardodev: @aidenybai @GuiBibeau

playground/components/editor.tsx
ethanniser: totally agree will change

playground/components/editor.tsx
```diff
@@ -15,6 +17,38 @@ import { GridResizer } from './grid-resizer';
...
+export const useFrameworkSyncUrl = () => {
+  const router = useRouter();
+  const params = useSearchParams();
+  const [framework, setFramework] = useAtom(frameworkAtom);
+
+  // keep url in sync with atom
+  useEffect(() => {
+    router.push(`/${framework === 'react' ? '' : `?framework=${framework}`}`);
+  }, [framework]);
```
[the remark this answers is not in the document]
ethanniser: Agree will change

playground/components/framework-switcher.tsx
GuiBibeau: I agree preset makes a bit more sense then framework. But this is mostly semantics and not really an issue for this PR. We could make a follow up for it if we want to.

playground/package.json
```diff
@@ -24,7 +24,8 @@
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "tailwindcss": "3.3.2",
-    "typescript": "^4.9.5"
+    "typescript": "^4.9.5",
```
GuiBibeau, line 27: Typescript can be move de dev dependencies.

playground/components/editor.tsx
```diff
@@ -5,15 +5,47 @@ import {
...
 import { files as reactViteFiles } from '@/configurations/react-vite';
 import { files as nextjsFiles } from '@/configurations/nextjs';
-import { frameworkAtom } from '@/atoms/framework';
+import { frameworkAtom, zFrameworks, type Framework } from '@/atoms/framework';
 import { MonacoEditor } from './monaco-editor';
 import { GridResizer } from './grid-resizer';
 import { FrameworkSwitcher } from './framework-switcher';
-import type { Framework } from '@/types';
+
+export const useFrameworkSyncUrl = () => {
```
GuiBibeau, line 18: let's move to a hook folder.

GuiBibeau (commented):
left a few comments, so far so good!

ethanniser:
address the comments, if all looks good I think we can look to merge @GuiBibeau thanks!

GuiBibeau (approved):

PR #457, opened 2023-06-26
fix: allow importing both for and block + flaatten no props

aidenybai:

This PR fixes importing `{ For, block }` not transforming, and removes `__props` when there are no props.

**Status**

### DIFF
--- a/packages/compiler/react/transform.ts
+++ b/packages/compiler/react/transform.ts
@@ -227,12 +227,12 @@ export const transformComponent = (
 
   // We want to add a __props property for the original call props
   // TODO: refactor this probably
-  dynamics.data.push({
-    id: t.identifier('__props'),
-    value: params?.length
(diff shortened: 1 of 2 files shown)


packages/compiler/react/transform.ts
```diff
@@ -227,12 +227,12 @@ export const transformComponent = (
 
   // We want to add a __props property for the original call props
   // TODO: refactor this probably
-  dynamics.data.push({
-    id: t.identifier('__props'),
-    value: params?.length
-      ? (params[0] as t.Expression)
-      : t.objectExpression([]),
-  });
+  if (params?.length) {
```
poteboy, line 230:
```suggestion
  if (params?.length && t.isExpression(params[0])) {
```
[nits] this would be safer?

poteboy (approved):

Issue #372, opened 2023-05-28
Save and Share Playground


GuiBibeau:
**Description:**

Implement a feature that allows users to save their current work and share it with others. This might involve generating a unique URL that can be shared.

**Acceptance Criteria:**

- [ ] User can save their current work in the playground.
- [ ] The saved state can be shared via a unique URL.

PR #458, opened 2023-06-26
saving/loading from local storage

ethanniser:
first step on #372 

attempt to implement simple saving/loading from localStorage, its not really there but theres some progress

tried to go all in on jotai, mainly following the ["Todos example with atomFamily and localStorage" example from the jotai showcase](https://githubbox.com/pmndrs/jotai/tree/main/examples/todos_with_atomFamily)

the serialization seems to work fine, but as for loading in the files I think simply modifying the "files" prop of the SandpackProvider doesnt seem to work, I'm assuming well have to use useSandpack and update each file one by one

### DIFF
--- a/playground/atoms/framework.ts
+++ b/playground/atoms/framework.ts
@@ -1,7 +1,0 @@
-import { atom } from 'jotai';
-import { z } from 'zod';
-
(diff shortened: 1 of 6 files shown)


tobySolutions:
Hmm, good work @ethanniser! I love this!

Issue #392, opened 2023-06-10
feat: Improve code to check values for null or undefined.


nojiritakeshi:
**Is your feature request related to a problem? Please describe.**
First of all, I apologize for my poor English.
This is a code improvement proposal, not a feature improvement.

**Describe the solution you'd like**
I think the following code section can be substituted with the code I propose.
`value === null || value === undefined`
My Proposal:
`value == null`
By using the 'equivalence operator', it should be possible to inspect only null and undefined together.

**Describe alternatives you've considered**
Existing Code
`value === null || value === undefined`
My Proposal:
`value == null`

**Additional context**

nojiritakeshi:
If the points I make are good ones, may I make a pull request?

SukkaW:
> If the points I make are good ones, may I make a pull request?

Currently, the eslint rules used by the repo are enforcing `value === null || value === undefined`. This is to prevent an edge case of `document.all`:

```js
document.all === undefined
// false
document.all === null
// false
document.all == null
// true
```

tobySolutions:
> > If the points I make are good ones, may I make a pull request?
> 
> Currently, the eslint rules used by the repo are enforcing `value === null || value === undefined`. This is to prevent an edge case of `document.all`:
> 
> ```js
> document.all === undefined
> // false
> document.all === null
> // false
> document.all == null
> // true
> ```
> 
> 


Hmm, interesting. Thanks Sukka!

SukkaW:
> Hmm, interesting. Thanks Sukka!

I am considering re-visit this. If `== null` is fast enough, we should disable that eslint rule (while nobody uses `docuemnt.all` now).

tobySolutions:
> > Hmm, interesting. Thanks Sukka!
> 
> I am considering re-visit this. If `== null` is fast enough, we should disable that eslint rule (since nobody uses `docuemnt.all` now).


I was gonna implement this fix in one of my PRs where ai was upgrading dependencies. I think I sort of lost track of it then.

tobySolutions:
> > Hmm, interesting. Thanks Sukka!
> 
> I am considering re-visit this. If `== null` is fast enough, we should disable that eslint rule (since nobody uses `docuemnt.all` now).

So, @nojiritakeshi can handle it then?

SukkaW:
> > I am considering re-visit this. If `== null` is fast enough, we should disable that eslint rule (since nobody uses `docuemnt.all` now).
> 
[...]

PR #485, opened 2023-07-17
fix: lots of stuff

### LINKED ISSUE #335: Searchbar should have a different placeholder within the blog
The placeholder for the search bar within the blog should be "Search blog..." instead of "Search documentation..."

Having the placeholder of the search bar within the blog be "Search blog..." will really help a lot.

Check out the newest article on the website's blog to see how this is: [Virtual DOM: Back in Block](https://millionjs.org/blog/virtual-dom)

Screenshot from 2023-05-04 15-10-38

### LINKED ISSUE #357: bug: ability to dedupe and use block creation in different ways
**Describe the bug**

Currently, the way to create blocks is very rigid, by definition of the compiler:

```jsx
const TestBlock = block(Test);
```

In here, `Test` must be used in a variable declaration. It can only be done once. Everything below will either produce unexpected behavior or throw an error:

```jsx
console.log(block(Test)); // not a var declaration
const TestBlock1 = block(Test);
const TestBlock2 = block(Test); // err, used twice
```

**Expected behavior**

This needs to be solved by the compiler. Basically, the hope is that it can extract out the block function into it's own variable declaration. Furthermore, Test should have a custom property that stores the block declarations for future reference.

```jsx
const extractedComponent = block(Test);

console.log(extractedComponent);
const TestBlock1 = block(Test);
Test.someProp = TestBlock1;
const TestBlock2 = Test.someProp
```

### LINKED ISSUE #425: feat: webpack.config.js example for millionjs which works with the official webpack doc?
**Is your feature request related to a problem? Please describe.**
When you go to 
https://webpack.js.org/configuration/plugins/
An example webpack.config.js file looks like this
```
module.exports = {
  //...
  plugins: [
    new webpack.DefinePlugin({
      // Definitions...
    }),
  ],
};
```
plugins are constructed. This is how it is in all of my projects too

When you go to 
https://million.dev/docs/install

example of the same config if you want to use million:
```
webpack.config.js
import million from 'million/compiler';
 
export default {
  plugins: [million.webpack()],
};
```

So now try to fit million in my project and it's instantly incompatible.
- million for me needs to be imported with require, cant use import
- million.webpack() is not a constructor, I can not do new million.webpack()
- even putting million in the plugin array, it fails to run
```
myProject\node_modules\million\node_modules\@babel\core\lib\config\full.js:266
    throw new _configError.default([`Preset ${formattedPresetName} requires a filename to be set when babel is called directly,`, `\`\`\``, `babel.transformSync(code, { filename: 'file.ts', presets: [${formattedPresetName}] });`
, `\`\`\``, `See https://babeljs.io/docs/en/options#filename for more information.`].join("\n"));
          ^
```

**Describe the solution you'd like**
I would like to use million in my project, but cant find a single project or config example where its used with react+typescript+webpack+babel
Can someone enlighten me? Can this work combined with those things and if so - how do I set it up?
How would a functional complete webpack.config.js file look like in a real demo project?

I am really confused as to how to setup million in the context of webpack with the documentation at the moment. Can we clarify if a standard webpack config would work and if so how? If thats not the case - can we get a single demo project that uses webpack? Guess my request here is just better documentation/example, but really the hope for me is I can fit it in my existing project to evaluate performance gains

### LINKED ISSUE #465: bug: In some cases, styles do not work well in blocks.
**Describe the bug**
The following will not work in block.
1. when style is set to a property that contains uppercase letters.
2. When you want to render processed arguments.

Specifically, this is the case when.
```typescript
export const RowBlock = block((adjective: string) => {
  const fuga = adjective + "abc";
  return (
    <tr>
      <td style={{ [adjective]: "red" }}>{fuga}</td>
```

**To Reproduce**
**Expected behavior**
That all will be well.

**Screenshots**
```typescript
// adjective === backgroundColor
export const RowBlock = block(({ adjective, color, noun }: RowProps) => {
  return (
    <tr style={{ backgroundColor: "gray" }}>
      <td style={{ [adjective]: "red" }}>{adjective}</td>
      <td>{color}</td>
      <td>{noun}</td>
    </tr>
  );
});
```
キャプチャ

**Device (please complete the following information):**
 - OS Windows10
 - Browser chrome
 - Version 114.0.5735.199

**Additional context**
I'm currently working on this issue.
I'll do my best to give you a good report.

### LINKED ISSUE #469: bug: using `ref` variable name when referencing DOM node with `useRef` leads to error
**Describe the bug**
Naming variable `ref` when referencing DOM node with `useRef`:
```js
const ref = React.useRef(null);
```
will lead to error:
```js
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

Check the render method of `_anonymous$2`.
    at MillionBlock (https://vitejsvitetmqweb-ayit--5173--e809191e.local-credentialless.webcontainer.io/node_modules/.vite/deps/million_react.js?v=b5dfed3d:823:42)
    at _anonymous$2 (https://vitejsvitetmqweb-ayit--5173--e809191e.local-credentialless.webcontainer.io/src/App.jsx?t=1688610481787:7:21)
    at App
```

Renaming the variable helps avoid the error: 
```js
const inputRef = React.useRef(null);
```

But `inputRef` will never reference DOM node. It always in it's initial state with `null` value.
The bug doesn't reproduce if you reference something else other than DOM node.

**To Reproduce**
Here's an example: https://stackblitz.com/edit/vitejs-vite-tmqweb?file=src%2FApp.jsx

**Device (please complete the following information):**
 - OS: Linux 64-bit
 - Browser: Chrome 114.0.5735.198 (Official Build) (64-bit)
 - Version: Million 2.4.9, React 18.2.0, Vite 4.3.9

### LINKED ISSUE #482: bug: Macro doc misplaced
**Describe the bug**

**To Reproduce**
1. Go to 'Million.dev'
2. Click on '/docs'
3. Scroll down to 'API Reference > below Internals'

**Expected behavior**
To be inside the [internals/macro.mdx] folder

**Screenshots**
Screenshot 2023-07-14 at 14 00 02

**Device (please complete the following information):**
 - OS [e.g. iOS]
 - Browser [e.g. chrome, safari] Chrome
 - Version [e.g. 22] Latest you can get

**Additional context**

aidenybai:

This PR makes several major fixes/changes (apologies for the MASSIVE pr):

- Added `babel-plugin-transform-react-createelement-to-jsx`, converts `React.createElement` calls to JSX, so we can parse it. Helpful if we ever need to parse libraries
- Fix #425 – pass filename into babel `transformAsync` function
- We no longer cache imports since it causes builds / dev to fail after change
- Allow users to hook into analytics (undocumented)
- Fix #357 
- Fix #465 
- Fixed `<For />` for virtualization
- Implement #392 
- Allow `as` prop for `block` and `<For />` for custom element
- Fix #469 
- Fix #335
- Fix #482
- Remove slots for blocks within blocks

**Status**


packages/compiler/react/transform.ts
```diff
@@ -511,20 +549,109 @@ export const transformJSX = (
...
+              jsxPath,
+            );
+          }
+
+          const value = property.value;
+          if (t.isBooleanLiteral(value) || t.isNullLiteral(value)) {
+            if (t.isNullLiteral(value) || !value.value) {
+              properties.splice(i, 1);
+              i--;
+              j--;
```
QuiiBz, line 598: Is that intended? If yes, we could probably do this to save a few bytes:
```suggestion
              i -= 2;
```

packages/compiler/react/transform.ts
```diff
@@ -511,20 +549,109 @@ export const transformJSX = (
...
+          styleObject.properties
+            .map((property) => {
+              if (t.isObjectProperty(property)) {
+                const value = property.value;
+                const key = property.key;
+                if (t.isIdentifier(key) && t.isLiteral(value)) {
+                  let kebabKey = '';
+                  for (let i = 0, j = key.name.length; i < j; ++i) {
+                    const char = key.name.charCodeAt(i);
+                    if (char < 97) {
```
QuiiBz, line 633: Maybe worth it to extract `97` to a top-level const with a proper name, because it's hard to understand what's going on

packages/million/block.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable eqeqeq */
```
QuiiBz, line 1: Do we want to disable this globally?

tobySolutions (approved):
LGTM!

PR #523, opened 2023-08-12
Use vite plugin hook to avoid double Babel pass

ArnaudBarre:
Hi! I've not yet fully tested it on a real codebase, but HMR is broken with the current setup so I wanted to dive into the code to help.

There are three things we can address:
- When splitting the component in two, the component is first named `_App2` and then renamed to `App` before export.
The issue is that the fast-refresh transformation register the component with the name "_App2" which is then at runtime excluded from being a component [here](https://github.com/facebook/react/blob/main/packages/react-refresh/src/ReactFreshBabelPlugin.js#L43). I didn't find what causses the behaviour in the babel plugin code, but if you could tell babel to use underscore suffix it would be great.
- After the file is transformed, Vite discover the `million/react` dep which was not in source code, generating a new deps optimisation and a page reload. We can hint Vite to always include this dep to avoid that
- Lastly, it would be great if both babel plugins (million and fast-refresh) could run at the same time. I not sure if that's possible and the first test are not successful when running million before fast-refresh in the same pass. In the other order (requires some hack), this seems to produce a valide output but still fails to HMR because the fast-refresh runtime still sees the functions as named "_App2"

Hope this can help! I think these issues should be fixed for the plugin to be adopted in large codebases, because running Babel twice instead of SWC ones can be a big DX cost with hundreds of components.


packages/compiler/plugin.ts
```diff
@@ -57,11 +61,13 @@ export const intro = (options: Options) => {
   }
 };
 
-export const unplugin = createUnplugin((options: Options = {}) => {
+// eslint-disable-next-line @typescript-eslint/default-param-last
+export const unplugin = createUnplugin((options: Options = {}, meta) => {
```
SukkaW, line 65: IMHO you shouldn't disable the `@typescript-eslint/default-param-last` rule here. Please provide a default value for `meta`.

SukkaW (changes requested):

packages/compiler/plugin.ts
ArnaudBarre: The meta argument will be injected by unplugin so it's never undefined.
I agree with the goal of the rule when creating API but here we don't control the API.
I can change to this to avoid the default on first argument:
```suggestion
export const unplugin = createUnplugin((_options?: Options, meta) => {
  const options = _options ?? {}
```

aidenybai (approved):

aidenybai:
@ArnaudBarre repinging if there are any updates here

ArnaudBarre:
So no updates on my part for now, this work is totally incomplete because it first requires to fix the compiler plugin so that intermediate component are not prefixed with underscore. Then I could look into how to make the one babel-pass compatible. I have a lot of works in the coming weeks so I don't the time right now to dive in the compiler code, the PR is just to get started on making the plugin works well in dev env.
Feel free to close if this is not a priority for you

aidenybai:
@ArnaudBarre no worries. I think this exploration is very interesting. I would be happy to fund the development of this issue if you'd like :)

ArnaudBarre:
I think #587 have fixed the prefix the issue.
I've done the error to create the branch PR from master, so a fork update delete the PR, will open another is the one pass feat if it works

ArnaudBarre:
Ok this not working, both plugins colide and generate undefined variables.
Only think you can keep from this PR:
```js
vite: {
  config() {
    return { optimizeDeps: { include: ['million/react'] } };
  },
}
```
So that people using only mode auto don't have initial deps re-optimize.

PR #537, opened 2023-08-17
version on react + run prettier

### LINKED ISSUE #534: [$35] Support React 16 + 17
### Describe the feature

Currently Million.js supports React 18. It would be nice to support React 16/17 builds. 

### Additional information

- [X] Would you be willing to help implement this feature?

rishi-raj-jain:
fix #534
/claim #534

**Status**

### DIFF
--- a/packages/react/utils.ts
+++ b/packages/react/utils.ts
@@ -1,8 +1,6 @@
-import { Fragment, createElement, isValidElement } from 'react';
-import { createRoot } from 'react-dom/client';
+import { Fragment, createElement, isValidElement, version } from 'react';
 import { REACT_ROOT, REGISTRY, RENDER_SCOPE } from './constants';
 import type { ComponentProps, ReactNode, Ref } from 'react';
-import type { Root } from 'react-dom/client';
 import type { VNode } from '../million';
 
 // TODO: access perf impact of this
@@ -45,11 +43,18 @@ export const renderReactScope = (vnode: ReactNode, unstable?: boolean) => {
   }
 
   const scope = (el: HTMLElement | null) => {
+    let root;
     const parent = el ?? document.createElement(RENDER_SCOPE);
-    const root =
-      REACT_ROOT in parent
-        ? (parent[REACT_ROOT] as Root)
-        : (parent[REACT_ROOT] = createRoot(parent));
+    if (version.startsWith('18')) {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const { createRoot } = require('react-dom/client');
+      root =
+        REACT_ROOT in parent
+          ? parent[REACT_ROOT]
+          : (parent[REACT_ROOT] = createRoot(parent));
+    } else {
+      root = parent[REACT_ROOT];
+    }
     root.render(vnode);
     return parent;
   };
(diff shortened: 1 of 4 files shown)


rishi-raj-jain:
So found that the errors in react 16 & 17 were of using createRoot (as it's not available in those versions).

Tested with all the versions, react 16, 17 and 18 with a vite project, by replacing the node_modules/million/dist with the locally built one (by pnpm build)

rishi-raj-jain:
@aidenybai 

Do I need to do anymore changes?

aidenybai:
good work!

packages/react/utils.ts
```diff
@@ -45,12 +43,25 @@ export const renderReactScope = (vnode: ReactNode, unstable?: boolean) => {
...
   const scope = (el: HTMLElement | null) => {
+    let root;
     const parent = el ?? document.createElement(RENDER_SCOPE);
-    const root =
-      REACT_ROOT in parent
-        ? (parent[REACT_ROOT] as Root)
-        : (parent[REACT_ROOT] = createRoot(parent));
-    root.render(vnode);
+    if (version.startsWith('18')) {
+      import('react-dom/client')
```
AbhiPrasad, line 49: This might break Webpack which will attempt to import `react-dom/client` even if using react 17 because bundlers are not smart enough to evaluate `version.startsWith('18')`.

I tested with and got the following error on https://github.com/getsentry/sentry (which uses React 17).

```
ERROR in ../node_modules/million/dist/chunks/utils.mjs 46:6-32
Module not found: Error: Can't resolve 'react-dom/client' in '/Users/abhijeetprasad/workspace/sentry/node_modules/million/dist/chunks'
resolve 'react-dom/client' in '/Users/abhijeetprasad/workspace/sentry/node_modules/million/dist/chunks'
  Parsed request is a module
  using description file: /Users/abhijeetprasad/workspace/sentry/node_modules/million/package.json (relative path: ./dist/chunks)
```

IMO the more holistic solution is to actually not do this built time - but instead split up the `renderReactScope` function into `renderReactScopev17andbelow` and `renderReactScope` and inject different one's depending on what version of react you're using, so we take care of this at built time instead of run time.

AbhiPrasad: I also tried adding million to https://github.com/romansndlr/react-vite-realworld-example-app with this patch (vite app using React 17) and things broke during build as well.

```
[vite]: Rollup failed to resolve import "react-dom/client" from ".yalc/million/dist/chunks/utils.mjs".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
error during build:
Error: [vite]: Rollup failed to resolve import "react-dom/client" from ".yalc/million/dist/chunks/utils.mjs".
This is most likely unintended because it can break your application at runtime.
```

`.yalc` comes from https://github.com/wclr/yalc - which is what I used to test million locally.

rishi-raj-jain: I think million new version containing this fix is yet to be released. Can you try replacing the dist folder with the dist folder from this?

AbhiPrasad: Might be worthwhile to also test Next 12 - which many folks are still using because they can't upgrade to Next 13 yet, and Next 12 uses React 17 by default.

Next.js and it's webpack setup might also cause issues here.

AbhiPrasad: > I think million new version containing this fix is yet to be released. Can you try replacing the dist folder with the dist folder from this?

I'm using the most recent version on master branch! I built the repo locally and directly imported it into my project.

rishi-raj-jain: Also, appreciate the feedback and you reaching out!

PR #545, opened 2023-08-20
hotfix: contexts

aidenybai:

This PR fixes contexts by reimplementing the root system to portal based system.

**Status**


AbhiPrasad (approved):
Two things that come to mind are:
1. This approach might break SSR.
2. This introduces the possibility of nested portals right? Do we need to avoid that scenario?

Otherwise generally this looks good, hard for me to examine in more detail without more test apps though.

AbhiPrasad:
Is there a way we can feature flag this to make it easier to test? Like we switch to the portal approach with an opt-in flag, ask people to test and then switch to this once we have more confidence.

oliverloops (approved):
I reviewed the core block re-implementation and see everything fine, I just have a couple of questions around this portal based system:

- What's the main purpose or migrate from the old one?
- This could make harder to make it compatible with 3rd party UI libraries and the future cross file blocks feature?

aidenybai:
going to run some beta tests

PR #575, opened 2023-09-12
 Enhancing Performance with React Concurrent Mode's `startTransition` 🚀

sanjaiyan-dev:
This PR brings minor performance enhancements by replacing the use of `queueMicrotask` with `startTransition` from React Concurrent mode. The advantages of using `startTransition` include:

1. **Improved Responsiveness:** `startTransition` prioritizes user interactions, leading to a smoother and more responsive user experience.

2. **Enhanced Control:** It offers better control over when updates are processed, allowing for optimized rendering.

3. **Concurrency Features:** Leveraging React Concurrent mode features like time-slicing can prevent performance issues and improve overall app performance.

Please review the PR, and don't hesitate to provide feedback if you spot any errors or have any suggestions for further improvements 🙏🏽

### DIFF
--- a/packages/react/block.ts
+++ b/packages/react/block.ts
@@ -1,11 +1,10 @@
-import { createElement, Fragment, useCallback, useMemo, useRef } from 'react';
+import { createElement, Fragment, useCallback, useMemo, useRef, startTransition } from 'react';
 import {
   block as createBlock,
   mount$,
   patch as patchBlock,
 } from '../million/block';
 import { MapSet$, MapHas$ } from '../million/constants';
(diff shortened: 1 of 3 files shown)


aidenybai:
this won't have any improvement because the mount/patch process isn't running in a react fiber scope

sanjaiyan-dev:
> this won't have any improvement because the mount/patch process isn't running in a react fiber scope

"Hello @aidenybai , I have a question regarding the use of `requestAnimationFrame` instead of `queueMicrotask`, especially in a function dedicated to rendering tasks. I apologize if this question seems basic.

PR #583, opened 2023-09-15
feat: copy sentry telemetry

aidenybai:


aidenybai:
@AbhiPrasad not sure if i'm doing this correctly, can you verify?

PR #584, opened 2023-09-17
feat: full hydration support

aidenybai:
This PR implements shallow hydration for Million.js **[Note: This is a big work in progress + is very messy]**

### Problem

Currently, Million.js doesn't properly hydrate. Let's say that `<Foo />` is a block component. 

```js
function Foo() {
  return <button onClick={() => alert('hello')}>Foo</button>
}
```

If the SSRed HTML is:

```html
<div>
  <button>Foo</button>
</div>
```

Then during hydration, React will hydrate the component:

```html
<div>
  <button>Foo</button>
   ^ now has event listener
</div>
```

Then, on next re-render, Million.js will `replaceChild` the `<button>` element

```html
<div>
	<slot><button>Foo</button></slot>
</div>
```

This is effectively a destructive method to approaching hydration.

### Solution

Instead of a full replace, we must reimplement the hydration from bottom up. This means, the vdom runtime, the react runtime, the react-server runtime and the compiler must be changed.

### PR Details 

**Status**

- [x] This PR **only** includes non-code changes, like changes to documentation, README, etc.


packages/kitchen-sink/src/main.tsx
```diff
@@ -9,19 +8,12 @@ import {
 } from 'react';
 import './style.css';
 import { ErrorBoundary } from 'react-error-boundary';
+import { FiberProvider } from 'its-fine';
+import { createRoot } from 'react-dom/client';
```
rishi-raj-jain, line 12: this'd again break for react v < 18

Drex72:
@tobySolutions Any issues with this pr?

PR #693, opened 2023-10-06
speech_detecting_build using Javascript

Bharath-KumarReddy:
When user speaks , some thing his speech is translated into words and that words are displayed dynamically on the notebook which is build with beautiful UI

once check it and merge it , kindly inform incase of further modifications ; _hacktoberfest_23;


tobySolutions (commented):
Hey there, I think your contributions were meant to be targeted at the Kitchen Sink, but, it doesn't seem like this matches at all.

Thank you.

PR #826, opened 2023-11-09
Fix for Auto Response Action

haideralipunjabi:
As per this [documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) `pull_request_target` is used for: 

> This event allows your workflow to do things like label or comment on pull requests from forks.


tobySolutions:
Thanks @haideralipunjabi!

PR #833, opened 2023-11-15
fix/#680/input value does not update on screen

### LINKED ISSUE #680: [$50] modified state value is not reflected into the input field.
### What version of `million` are you using?

2.6.4

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

yarn

### What operating system are you using?

Windows

### What browser are you using?

Chrome

### Describe the Bug

The variable's state get reset when passing empty string but the input field doesn't get emptied. Problem is demonstrated more clearly through a video

https://github.com/aidenybai/million/assets/12785891/abd5b0cc-39e6-4271-bb73-d09643629e80

### What's the expected result?

When resetting the status to empty value, then input field should be reset as well.

### Link to Minimal Reproducible Example

Check the video (easy to reproduce)

### Participation

- [ ] I am willing to submit a pull request for this issue.

ologbonowiwi:
- test: ensure that value is being cleaned
- fix: set value explicitly to input, select, or text area elements

This fixes #680. The issue is that we changed the attribute via `setAttribute, ' which does not re-render/update the DOM value after it was already set (which is different from setting explicitly using `element.value = 'something'`). This is mentioned on [MDN's setAttribute documentation](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute#gecko_notes)
> Using `setAttribute()` to modify certain attributes, most notably value in XUL, works inconsistently, as the attribute 
specifies the default value. To access or modify the current values, you should use the properties. For example, use 
Element.value instead of Element.setAttribute().

Here is a JSFiddle showing the behavior happening: https://jsfiddle.net/hybrk97u/

This fix introduces a manual value set for `HTMLInputElement`, `HTMLSelectElement`, and `HTMLTextAreaElement`. These are the only elements that suffer from this issue from what I found here.

**Status**

- [X] Code changes have been tested against prettier, or there are no code changes
- [X] I know how to update typings and have done so, or typings don't need updating

- [X] This PR changes the codebase
  - [X] This PR changes the internal workings with no modifications to the external API (bug fixes, performance 
improvements)

/claim #680

### DIFF
--- a/packages/million/dom.ts
+++ b/packages/million/dom.ts
@@ -227,4 +227,13 @@ export const setAttribute = (
   } else {
     removeAttribute$.call(el, name);
   }
+
+  const isInput = el instanceof HTMLInputElement;
+  const isSelect = el instanceof HTMLSelectElement;
+  const isTextArea = el instanceof HTMLTextAreaElement;
+
+  if (name === 'value' && (isInput || isSelect || isTextArea)) {
+    setAttribute$.call(el, name, String(value));
+    el.value = value as string;
+  }
 };
--- a/test/block.test.ts
+++ b/test/block.test.ts
@@ -101,4 +101,31 @@ describe.concurrent('block', () => {
       '<div><h1>Hello</h1> World<p title="baz" style="margin: 1px;" class="bar"></p></div>',
     );
   });
+
+  it('should clear input inside block if the value is empty', ({ expect }) => {
+    const block = createBlock((props?: MillionProps) => {
+      return {
+        type: 'div',
+        props: {
+          children: [
+            {
+              type: 'input',
+              props: {
+                value: props?.foo,
+              },
+            },
+          ],
+        },
+      };
+    });
+
+    const main = block({ foo: 'foo' });
+
+    main.m();
+    expect(main.l?.outerHTML).toEqual('<div><input value="foo"></div>');
+
+    main.p(block({ foo: '' }));
+
+    expect(main.l?.outerHTML).toEqual('<div><input value=""></div>');
+  })
 });
--- a/test/block.test.ts
+++ b/test/block.test.ts
@@ -127,5 +127,5 @@ describe.concurrent('block', () => {
     main.p(block({ foo: '' }));
 
     expect(main.l?.outerHTML).toEqual('<div><input value=""></div>');
-  })
+  });
 });


packages/million/dom.ts
```diff
@@ -227,4 +227,13 @@ export const setAttribute = (
   } else {
     removeAttribute$.call(el, name);
   }
+
+  const isInput = el instanceof HTMLInputElement;
+  const isSelect = el instanceof HTMLSelectElement;
+  const isTextArea = el instanceof HTMLTextAreaElement;
+
+  if (name === 'value' && (isInput || isSelect || isTextArea)) {
+    setAttribute$.call(el, name, String(value));
```
ologbonowiwi, line 236: This is unnecessary, but it can show different values between the `value` attribute and the `value` property. I'm letting it here for consistency and to avoid user confusion.

test/block.test.ts
```diff
@@ -101,4 +101,31 @@ describe.concurrent('block', () => {
...
+    });
+
+    const main = block({ foo: 'foo' });
+
+    main.m();
+    expect(main.l?.outerHTML).toEqual('<div><input value="foo"></div>');
+
+    main.p(block({ foo: '' }));
+
+    expect(main.l?.outerHTML).toEqual('<div><input value=""></div>');
```
ologbonowiwi, line 129: this was breaking before the fix. the return here was `<div><input></div>` instead of the `<div><input value=""></div>` expected.

packages/million/dom.ts
```diff
@@ -227,4 +227,13 @@ export const setAttribute = (
   } else {
     removeAttribute$.call(el, name);
   }
+
+  const isInput = el instanceof HTMLInputElement;
+  const isSelect = el instanceof HTMLSelectElement;
+  const isTextArea = el instanceof HTMLTextAreaElement;
+
+  if (name === 'value' && (isInput || isSelect || isTextArea)) {
```
jkcs, line 235: @ologbonowiwi I don't think we can directly use this to determine the update of the value attribute. For example, the `<option>` tag also has a value attribute. Here, I suggest using `el.hasOwnProperty('value')` or, even more boldly
```js
if (el.hasOwnProperty(name)) {
    el[name] = value;
}
```
What do you think? ( I am not the maintainer.)

ologbonowiwi: you're right, but usually, we update `option` element values through `select` instead of via `option` itself. 

we indeed can have more cases of this issue, but we can wait to have more issues and then add to the conditions rather than over-optimize something that solves the original issue :eyes:

let me know your thoughts

jkcs: I think you're right. I'm wondering if there's a method that can consider all the cases. Currently, I don't have any good ideas either. Besides the `value` property, I also realized that the `textContent` property might have the same issue.

tobySolutions:
Hmm, good work @ologbonowiwi. Thanks.

ologbonowiwi:
Thanks @tobySolutions! Would you mind review it?

I can collect the bounty only when the PR is merged.

ologbonowiwi:
Hey @tobySolutions @aidenybai can you review this one guys?

sdegueldre:
This fix is incomplete: the `checked` attribute on inputs of type checkbox/radio also needs the JS property set on the underlying DOM node. It's also worth noting that in React, you can pass the prop `defaultValue` or `defaultChecked` for uncontrolled components, which will only set the attribute but not the property, so this might need to account for those too for compatibility.

ologbonowiwi:
Can you provide a jsfiddle exemplifying it @sdegueldre?

I couldn't reproduce here.

skushagra9:
Is this issue solved compeletely ?
If not please assign me

0x15f:
If there is anything I can do to assist please let me know, this PR solves massive blockers for using Million in various projects.

Aslemammad:
Any progress on this? I'd like to have a status check on this PR!

PR #841, opened 2023-11-27
YAY 3.0!!!

aidenybai:
WEW DID TI!!!


packages/compiler/babel.ts
```diff
@@ -0,0 +1,166 @@
...
+        enter(programPath: NodePath<t.Program>) {
+          info = {
+            imports: state,
+            programPath,
+            blocks: new Map<string, t.Identifier>(),
+            filename,
+            options,
+          };
+
+          programPath.traverse<Info['imports']>(
```
lxsmnsyc, line 62: you probably don't need to do this as a separate traverse call.

packages/compiler/babel.ts
```diff
@@ -0,0 +1,166 @@
...
+            {
+              ImportDeclaration(path, state) {
+                const importDeclaration = path.node;
+                const importSource = importDeclaration.source;
+                if (!importSource.value.includes('million')) return;
```
lxsmnsyc, line 67: Wait how accurate is this or probably the other question is, what patterns are you trying to check against

packages/compiler/babel.ts
```diff
@@ -0,0 +1,166 @@
...
+                 */
+                const source = resolveImportSource(options, importSource.value);
+
+                for (
+                  let i = 0, j = importDeclaration.specifiers.length;
+                  i < j;
+                  ++i
+                ) {
+                  const specifier = importDeclaration.specifiers[i];
+                  if (
```
lxsmnsyc, line 82: I guess no namespace support yet? (i.e. `import * as x from 'x'`)

packages/compiler/babel.ts
```diff
@@ -0,0 +1,166 @@
...
+                    case 'block':
+                    case 'For':
+                    case 'renderReactScope':
+                      if (!state[name]) {
+                        state[name] = specifier.local.name;
+                      }
+                      if (!state.aliases[name]) {
+                        state.aliases[name] = new Set();
+                      }
+                      state.aliases[name]?.add(specifier.local.name);
```
lxsmnsyc, line 101: unnecessary optional chaining

packages/compiler/auto.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  const isInProgramScope =
+    t.isIdentifier(rawComponent.id) &&
+    info.programPath.scope.hasOwnBinding(rawComponent.id.name);
+
+  if (!isInProgramScope) return;
+
+  if (
+    t.isCallExpression(path.parent) &&
+    t.isIdentifier(path.parent.callee, { name: 'block' }) // probably need a import level check later
```
lxsmnsyc, line 29: you definitely should 🤣

packages/compiler/auto.ts
```diff
@@ -0,0 +1,306 @@
...
+    const hasUseClientDirective = directives.some((directive) => {
+      return directive.value.value === 'use client';
+    });
+    if (!hasUseClientDirective) return;
+  }
+
+  if (t.isVariableDeclarator(rawComponent)) {
+    if (
+      !t.isArrowFunctionExpression(rawComponent.init) &&
+      !t.isFunctionExpression(rawComponent.init)
```
lxsmnsyc, line 50: Tip for the future: cover cases where expression is wrapped in a TS expression (e.g. `x as MyType` which is the same as `x`)

packages/compiler/auto.ts
```diff
@@ -0,0 +1,306 @@
...
+  } else {
+    componentPath = path as NodePath<t.FunctionDeclaration>;
+  }
+
+  if (!t.isIdentifier(rawComponent.id)) return;
+  if (!isComponent(rawComponent.id.name)) return;
+
+  const globalPath = path.findParent(
+    (path) => path.parentPath?.isProgram() || path.isProgram()
+  )!;
```
lxsmnsyc, line 75: I believe `path.getProgramParent()` exists

packages/compiler/auto.ts
```diff
@@ -0,0 +1,306 @@
...
+  };
+
+  const skipIds: (string | RegExp)[] = [];
+
+  if (typeof info.options.auto === 'object' && info.options.auto.skip) {
+    skipIds.push(...info.options.auto.skip);
+  }
+
+  blockStatementPath.traverse({
+    Identifier(path) {
```
lxsmnsyc, line 120: Be careful on this one, as this covers non-expression identifiers like object keys. I would recommend traversing `Expression` instead and then checking if the node is an `Identifier`

packages/compiler/auto.ts
```diff
@@ -0,0 +1,306 @@
...
+      componentInfo.components++;
+      path.skip();
+    },
+    JSXExpressionContainer(path) {
+      if (t.isJSXEmptyExpression(path.node.expression)) return;
+
+      if (
+        t.isCallExpression(path.node.expression) &&
+        t.isMemberExpression(path.node.expression.callee) &&
+        t.isIdentifier(path.node.expression.callee.property, { name: 'map' })
```
lxsmnsyc, line 172: This has holes, but whatever, could be improved. (Same goes to other bruteforce checks)

packages/compiler/auto.ts
```diff
@@ -0,0 +1,306 @@
...
+        path.stop();
+        return;
+      }
+    },
+  });
+
+  const good =
+    componentInfo.elements + componentInfo.attributes + componentInfo.text;
+  const bad = componentInfo.components + componentInfo.expressions;
+  const improvement = (good - bad) / (good + bad);
```
lxsmnsyc, line 217: any paper for this? 🤓

packages/compiler/block.ts
```diff
@@ -1,51 +1,288 @@
...
+    t.isVariableDeclarator(callExpressionPath.parent) &&
+    t.isIdentifier(callExpressionPath.parent.id)
+      ? callExpressionPath.parent.id
+      : null;
+
+  if (
+    !t.isIdentifier(rawComponent) &&
+    !t.isFunctionDeclaration(rawComponent) &&
+    !t.isFunctionExpression(rawComponent) &&
+    !t.isArrowFunctionExpression(rawComponent)
```
lxsmnsyc, line 62: Same case here for nested expressions

packages/compiler/block.ts
```diff
@@ -1,51 +1,288 @@
...
+        : publicComponent?.name || 'Anonymous';
+    const anonymousComponentId = getUniqueId(callExpressionPath, name);
+
+    /**
+     * Replaces function expression with identifier
+     * `block(() => ...)` to `block(anonymous)`
+     */
+
+    // TODO: check this
+    const paths = callExpressionPath.parentPath.parentPath?.insertBefore(
```
lxsmnsyc, line 146: I'm not sure if direct access to `parentPath` is safe here. I'd usually crawl towards the parent statement path and then insert it before. Needs testing.

lxsmnsyc (commented):
Summary (and tips):
- There's a lot of node type checks however in some cases they may or may not be accurate. I'm not sure if `million` currently intends to run only on JS code or before TS compilation but that's the part of the compiler that can be improved.
- Probably the import identifier checks. If there's any other form of "identifier" checks, you should also need to check if the "identifier" you are testing against is the same identifier that is actually being referenced.

Note: I have 0 idea yet on how Million expects a code to look and how it should look like after so my review is kinda limited and I just applied based on what I usually do on my other Babel-based libraries

or just hire me to fix these

Issue #567, opened 2023-09-02
[$500] Remove slots


aidenybai:
### Describe the feature

Remove the usage of <slot> elements from everything-from compile to runtime

### Additional information

- [X] Would you be willing to help implement this feature?

feliciien:
https://discord.com/channels/938129049539186758/1148796877853163530

rishi-raj-jain:
@feliciien We need the <slot> to be removed from components as it simply blocks all the parent CSS properties.

feliciien:
I'm no longer authorized to work on this 😅

rishi-raj-jain:
Oh okay okay

rishi-raj-jain:
Why's that?

itsvanshchavda:
Hello brother can you give some more description about that issue I want to work on this

ogil7190:
@aidenybai is this issue open to solve?

tobySolutions:
> @aidenybai is this issue open to solve?

Yep! Go ahead! 💯💯

evoxf1:
is it still open ?

dennismzia:
> is it still open ?

seems like so try your luck man

joshpcausey:
@tobySolutions what element would should we render instead of <slot>

tobySolutions:
> @tobySolutions @aidenybai what element should we render instead of slot



cc @aidenybai

jeet-dhandha:
Can someone tell me why are we using slots to wrap components ?

devilserpent123:
is this solved?

tobySolutions:
> is this solved?
> 

Nope, not yet @devilserpent123. Do you want to take it on?

devilserpent123:
Yeah sure.

tobySolutions:
> Yeah sure.
> 

Awesome! Go ahead.

ologbonowiwi:
@tobySolutions @aidenybai, is there any component we should use instead of a slot, or should we render just the plain component received?

ologbonowiwi:
Can you assign this to me? I'll handle it. 

/claim #567

sankalp142002:
Hello, @tobySolutions. Can I take this issue forward if the work is still not done?

tobySolutions:
> Hello, @tobySolutions. Can I take this issue forward if the work is still not done?

Sure thing, go ahead @sankalp142002!

aidenybai:
@aslemammad  Jan 15

[...]

PR #845, opened 2023-12-19
:hammer: fix

japrozs:
/claim #567

### DIFF
--- a/packages/million/template.ts
+++ b/packages/million/template.ts
@@ -45,7 +45,7 @@ export const renderToTemplate = (
       i: [],
     });
 
-    return '<slot/>';
+    return '<></>';
   }
 
   let props = '';
--- a/packages/preact/constants.ts
+++ b/packages/preact/constants.ts
@@ -1,6 +1,6 @@
 import { useEffect } from 'preact/hooks';
 
-export const RENDER_SCOPE = 'slot';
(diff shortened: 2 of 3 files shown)


ologbonowiwi:
This doesn't work. Having a div will break the CSS selector, same as slot breaks currently.

tobySolutions:
Hmm, what do you think @ologbonowiwi, @japrozs?

ologbonowiwi:
I've tried this approach but we'll end having the same issue that we have now (the CSS selector will not account the additional `<div>`, as does not accommodate the additional `<slot>`)

rishi-raj-jain:
There shall be no component as far as I see. <select>s for example shall have no nested items inside them.

ologbonowiwi:
👆🏽 Exactly 

currently 
```
<div>
  // elements
</div>
```
becomes 
```
<slot>
  <div>
    // elements
  </div>
</slot>
```
but as @rishi-raj-jain said, we need to keep the original structure

tobySolutions:
Thanks @rishi-raj-jain, @ologbonowiwi; are you interested in setting up a meeting @ologbonowiwi to go over what your fix does?

skushagra9:
Is this issue solved compeletely ?
If not please assign me

PR #884, opened 2024-01-17
feat: 3.0 compiler

lxsmnsyc:
# 🚧 UNDER CONSTRUCTION 🚧

## Goals

- [x] Add proper import identifier detection
- [x] Add proper support for client-server compilation variance
- [x] Add descriptive name generation for compiler-generated blocks
- [x] Rework compiler-generated blocks' prop generation
- [x] Add internal API for comparing block props
- [x] Rework compiler-generated code for Portals
- [ ] Rework `For` handling
- [x] Rework `.map` handling
- [x] Add internal API for managing renderPortalScope
- [x] Add support for compiling all JSX owned by the block component (unlike only `return`)
- [x] Rework auto mode


lxsmnsyc (commented):
As you notice, most of the compiler-related files have a `.new.ts` here. There's two reasons:
- I'm still looking at how the old one works while not overwriting it with the new one.
- I needed a way to test both old and new compiler versions at the same time.

tobySolutions:
Amazing stuff!!! This is insane @lxsmnsyc!

lxsmnsyc:
One big change is that `block` components no longer compile into `block` but into `compiledBlock`. `compiledBlock` manages both hoisted templates and portals via statically analyzing how much of the elements needs to be rendered into a portal.

Another thing that this unlocks is that you can now do branching statements inside `block` properly.

tobySolutions:
> One big change is that `block` components no longer compile into `block` but into `compiledBlock`. `compiledBlock` manages both hoisted templates and portals via statically analyzing how much of the elements needs to be rendered into a portal.
> 
> Another thing that this unlocks is that you can now do branching statements inside `block` properly.

Oh damn! That sounds awesome; and by "branching statements" you mean conditionals, yes?

lxsmnsyc:
@tobySolutions yes. I believe the previous block compilation didn't support this, because of the compiler output for portals.

tobySolutions:
> @tobySolutions yes. I believe the previous block compilation didn't support this, because of the compiler output for portals.

Yes, you are right! Amazing work Alex, this is huge progress!

lxsmnsyc:
Current issues (as per kitchen-sink):
- Guestbook
  - State lagging (requires another re-render for the new item to render)
- Markdown
  - Missing ref
- Million Quiz
  - Cannot convert Symbol to string
- QR Code
  - Missing ref
- Task Tracker
  - Loses focus on state change when portal is defined as unstable
- Todo List
  - State Lagging (same with Guestbook)
- Type Race
  - Flickering when portal is defined as unstable

lxsmnsyc:
- `_c` has been disabled for now since there's no good metric for it to be enabled. It's tricky to know if a component is actually relying on some underlying React Fiber.
- `rsc` is yet to be supported.
- SSR needs to be handled. I'm lacking idea on that part but for now, `compiledBlock` is 1:1 code in both server and client.
- Lacks warning logs for now.
- Telemetry is to be implemented.

aidenybai (approved):

lxsmnsyc:
@Aslemammad @aidenybai I'm closing this PR down, but not the branch, so feel free to backup things that you wanted.

PR #938, opened 2024-02-03
fix: `"type": "module"` support

lxsmnsyc:


lxsmnsyc:
What's up with the build, I thought it is included in `unplugin`

PR #946, opened 2024-02-05
fix: unique `key` error

### LINKED ISSUE #942: Warning: Each child in a list should have a unique "key" prop.
### What version of `million` are you using?

3.0.2

### Are you using an SSR adapter? If so, which one?

None

### What package manager are you using?

bun, yarn

### What operating system are you using?

wsl

### What browser are you using?

edge, chrome

### Describe the Bug

create fresh vite using bun or yarn
install millionjs using bun or yarn

followed the instructions 

```ts
export default defineConfig({
  plugins: [million.vite({ auto: true }), react()],
});
```

then I do.. bun run dev.

look at the console.
 
image

### What's the expected result?

expected result is to show non of that error at all because its a fresh project. 

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/vitejs-vite-ya4ek3

### Participation

- [x] I am willing to submit a pull request for this issue.

lxsmnsyc:
Fixes the unintentional unique `key` error thrown by React because of `block`

- Fixes #942

### DIFF
--- a/packages/react/block.ts
+++ b/packages/react/block.ts
@@ -1,14 +1,14 @@
-import { createElement, Fragment, useCallback, useMemo, useRef } from 'react';
 import type { ComponentType, Ref } from 'react';
+import { Fragment, createElement, useCallback, useMemo, useRef } from 'react';
 import {
   block as createBlock,
   mount$,
   patch as patchBlock,
 } from '../million/block';
-import { MapSet$, MapHas$ } from '../million/constants';
-import type { Options, MillionProps, MillionPortal } from '../types';
+import { MapHas$, MapSet$ } from '../million/constants';
+import type { MillionPortal, MillionProps, Options } from '../types';
+import { Effect, REGISTRY, RENDER_SCOPE, SVG_RENDER_SCOPE } from './constants';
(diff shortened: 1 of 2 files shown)


packages/react/block.ts
```diff
@@ -65,17 +65,14 @@ export const block = <P extends MillionProps>(
...
     for (let i = 0; i < childrenSize; ++i) {
-      children[i + 2] = portalRef.current[i]?.portal;
+      children[i] = portalRef.current[i]?.portal;
     }
 
-    const vnode = createElement(Fragment, { children });
+    const vnode = createElement(Fragment, {}, marker, createElement(Effect, {
+      effect,
+      deps: hmrTimestamp ? [hmrTimestamp] : [],
+    }), children);
```
Aslemammad, line 75: is it possible that portals (children) won't have a key and the warning would be shown again?

What I'd suggest is that you spread the children by `, ...children`, so this way the whole error would be removed. 

```
}), ...children);
```

lxsmnsyc: spreading might hit the arguments bound because it has a lower bound than arrays.

PR #948, opened 2024-02-06
feat(experimental): remove slots for blocks

Aslemammad:
The only way to resolve this is to divide and conquer it! After 3 implementations, all of them failed and I don't think we can deal with this easily.

I decided to divide this feature/breaking-change into multiple parts and add them with the `experimental_options.noSlot` flag or the `experimental_noSlot` prop. 

In this PR, I remove slots from blocks only and it does not even support siblings yet (or maybe never) but that's a good start I think. 

We ship this to the users with those flags and see their reactions and responses! 

The way to try this is to do this:

```ts
import { experimental_options }} from 'million/experimental'

experimental_options.noSlot = true;

// or 

block(() => ..., { experimental_noSlot: true })
```

Now I kinda understand why React is doing what it's doing with the experimental releases.


tobySolutions:
Hmm, I like this idea 🤔🤔. I think it's a good start.

PR #954, opened 2024-02-08
fix: gatsby object object

Aslemammad:
Since portals were not being scoped the react element object was rendered as [object object] in the million mounts. I fixed this by returning some code + some modifications.

The other thing this PR does is that it tells folks to have their components in jsx/tsx. since one of the repros I had was using .js (gatsby) and the components were not being optimized.


tobySolutions:
> Since portals were not being scoped the react element object was rendered as [object object] in the million mounts. I fixed this by returning some code + some modifications.
> 
> The other thing this PR does is that it tells folks to have their components in jsx/tsx. since one of the repros I had was using .js (gatsby) and the components were not being optimized.

Hmm, nice work @Aslemammad. Thanks for this. Are you giving the user a compiler message so they are aware of that caveat (the .jsx/.tsx) thing or do we need to document that?

PR #1009, opened 2024-03-26
fix: `log` compiler option

rnln:
There was a confusing behavior caused by the strange logic in the compiler logging.
It is possible to turn off the logging only if the `log` option has been set to a truthy value, which is not equal to `'info'`, because of the condition `!ctx.options.log || ctx.options.log === 'info'`.
However, at the same time, the type of the `log` option is either `boolean | 'info' | undefined`, so there is no valid value with type that can turn off the logging.
I think that this PR will make the logic of the `log` option a little clearer.

**Status**

### DIFF
--- a/packages/compiler/auto.ts
+++ b/packages/compiler/auto.ts
@@ -205,12 +205,11 @@ function shouldTransform(
       : 0.1;
 
   if (improvement <= threshold) return false;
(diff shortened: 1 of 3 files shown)


lxsmnsyc:
Related (although yours is earlier): https://github.com/aidenybai/million/pull/1020/files

I'll probably merge this one and just tweak my other PR.

